PR #24, opened 2019-01-26
Fix inconsistent id naming of shoes

ScallyGames:


jrconway3:
This repo has been updated to sanderfrenken's fork. However, I won't close this PR yet. Its possible these same issues still persist, so I'll look into seeing if they do still persist before closing it.

jrconway3:
Okay, yeah, sanderfrenken's fork auto updates these ID's so this is definitely not needed anymore. Going forward, things will be a bit more standardized now that its on an org.

Issue #35, opened 2025-01-03
Licensing improvements


sanderfrenken:
- [x] Add license filter
- [x] Generate credits.csv from licenses info in json
- [x] Update readme on new license structure 
- [x] Disable merging of adding new assets without a valid license

jrconway3:
So this is all done now, right? I need to download the latest updates and check, the only things I could think might be left are the readme and disabling merging of assets without a valid license, neither of which should be that hard.

sanderfrenken:
yes the readme and disable merging is left. Will do that Friday

jrconway3:
I wasn't aware you could even do that...

sanderfrenken:
Disable merging is already in place, the generate-sources will throw an error now when no valid license is found for an asset

PR #44, opened 2025-01-04
Generate credits csv when creating sources

sanderfrenken:
- Generate credits file for all separate files on site generation
- Rename generate_html.js to generate_sources.js
- Remove old credits.csv
- Update readme

### DIFF
--- a/.github/workflows/validate-index_html.yml
+++ b/.github/workflows/validate-index_html.yml
@@ -19,7 +19,7 @@ jobs:
 
       - name: Validate HTML generation
         run: |
-          node scripts/generate_html.js
+          node scripts/generate_sources.js
       - name: Assert git diff
         run: |
           if [[ `git status --porcelain` ]]; then
--- a/README.md
+++ b/README.md
@@ -29,9 +29,8 @@ Either way, make sure this credits file is accessible from within your game or a
 
 **Important: all art submitted to this project must be available under the [Creative Commons Attribution-ShareAlike 3.0](http://creativecommons.org/licenses/by-sa/3.0/) and/or the [GNU GPL 3.0](http://www.gnu.org/licenses/gpl-3.0.html) licenses**.
 - **By contributing original work to this project, you agree to license any orginal contributions made by you under _both_ of these licenses.**
-- If you are submitting art that was made by (or derived from work made by) someone else, please be sure that you have the rights to distribute that art under one of the two licenses.
 
-When adding files to the project, please add a row to `CREDITS.csv` for each file you add. Note the entire list of authors for that image, a URL for each piece of art from which this image is derived, and a list of licenses under which the art is available.
+- If you are submitting art that was made by (or derived from work made by) someone else, please be sure that you have the rights to distribute that art under one of the two licenses.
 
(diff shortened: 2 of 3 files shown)


sanderfrenken:
Relates to https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/issues/35

Adding license filter in a separate PR later.

jrconway3:
As I stated in the other PR, I was against changing or renaming the credits.csv...

However, that only applies due to the fact that bluecarrot16 and other entries on OGA link back to that credits.csv. This is a new fork now and so the old credits.csv is still there, unless you update your fork to the latest changes here. So I think that's fine, but personally, I don't really see the reason why we can't just keep the credits.csv the same filename.

jrconway3:
Once I have a chance to test this I'll approve, though again, I still think we'd be better off keeping the credits.csv the same filename.

sanderfrenken:
> As I stated in the other PR, I was against changing or renaming the credits.csv...
> 
> However, that only applies due to the fact that bluecarrot16 and other entries on OGA link back to that credits.csv. This is a new fork now and so the old credits.csv is still there, unless you update your fork to the latest changes here. So I think that's fine, but personally, I don't really see the reason why we can't just keep the credits.csv the same filename.

Can you provide an example of such a link back? In general I think that is quite fragile. The credits file is now in the root, but that might also change in the future.

I renamed it in order to more clearly communicate the intention of this file (listing the credits of all files). And I also didn't like the UPPERCASE usage in its name ;)

sanderfrenken:
> Once I have a chance to test this I'll approve, though again, I still think we'd be better off keeping the credits.csv the same filename.

no rush, I don't see a real argument to keep the old not so descriptive name atm though. Prefer to make it more descriptive.

sanderfrenken:
Renamed it back to CREDITS.csv, if you really feel strong about it I can live with it :) Don't want to waste too much time on this and I am now starting with the credits filter.

jrconway3:
Apologies, I've been distracted the past few days. I'm not THAT strongly attached to it. That being said, I thought I had seen the credits.csv linked to on OGA in some releases.

Anyway, I'll take a look at the updates tonight.

sanderfrenken:
> Apologies, I've been distracted the past few days. I'm not THAT strongly attached to it. That being said, I thought I had seen the credits.csv linked to on OGA in some releases.
> 
> Anyway, I'll take a look at the updates tonight.

I checked a couple releases but couldn't find any reference, but maybe there is in some of them. I don't really understand why there would be a reference to it though.

Gaurav0:
I took a quick look and only found a reference to CREDITS.csv here: https://opengameart.org/content/lpc-clothing-updates

It can easily be updated by @bluecarrot16

jrconway3:
I thought I had seen more than that, might be others than just bluecarrot.

sanderfrenken:
Ah, well then let's just stick with CREDITS.csv. It's ok and as said, never change a winning team...

I also created a draft for license filtering. Needs some more polish here and there but I think functionally this is what it will be:

https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/pull/47

PR #51, opened 2025-01-07
add button to remove assets incompatible with licenses

Gaurav0:
Related Discussion: #50 

This adds a button to License filter section that will allow the user to remove any selected assets incompatible with licenses selected in the license filter.

~It will NOT remove the default selections if they are selected but incompatible. (But it will not select the defaults if they are not). Is this good?~

### DIFF
--- a/sources/chargen.js
+++ b/sources/chargen.js
@@ -5,6 +5,8 @@ $.expr[':'].icontains = function(a, i, m) {
       .indexOf(m[3].toUpperCase()) >= 0;
 };
 
+$.fn.reverse = [].reverse;
+
 $(document).ready(function() {
 
   var matchBodyColor = true;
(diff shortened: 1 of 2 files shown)


sanderfrenken:
Looks good! I am only thinking do we need that check for a default asset? 

Currently you set the `body-Body_color_light` (and corresponding head) asset to be a default, and as a result when selecting a non-compatible license with these two assets, the button to remove the incompatible assets will not show if I understand it correctly, right?

Is that necessary? The other color variants for the same asset (with the same license, eg `body-Body_color_amber`) will behave differently, as those are not marked as default.

Gaurav0:
No, its not necessary. I just thought getting a blank sheet looked weird. Removed.

sanderfrenken (approved):

PR #54, opened 2025-01-10
Changes to licensing information in readme

sanderfrenken:


README.md
```diff
@@ -11,9 +11,9 @@ The Liberated Pixel Effort is a collaborative effort from a number of different
 
 ### Licensing and Attribution (Credits)
 
-All art distributed with this project (all images in the `spritesheets` subdirectory) is licensed under the [GNU GPL 3.0](http://www.gnu.org/licenses/gpl-3.0.html) ([text](gpl-3_0.txt)) and/or [Creative Commons Attribution-ShareAlike 3.0](http://creativecommons.org/licenses/by-sa/3.0/) ([text](cc-by-sa-3_0.txt)) license(s). Some art may be available under other licenses too.
+All art distributed with this project (all images in the `spritesheets` subdirectory) is licensed under one or more license(s).
```
Gaurav0, line 14: Let's add the full list of licenses and links here.

Gaurav0 (changes requested):
Why are we deleting the license files?

Also, can you add a statement that the source code in the repo is GPL 3?

sanderfrenken:
> Why are we deleting the license files?

I find it redundant and even incomplete. There are assets that are licensed for example solely under CC0 for example.

But it's redundant either way (and the up to date license is available online). the full license info is generated in the textfield or accessible via the credits.csv. Moreover, the link to each license is now accessible via the license filters.

Its plain overhead imo to put the same information in the readme, which you will also need to keep up to date.

> Also, can you add a statement that the source code in the repo is GPL 3?

This is already covered by the LICENSE file in the root of the repo

README.md
sanderfrenken: please read my reply to the other comment

Gaurav0:
> > Why are we deleting the license files?
> 
> I find it redundant and even incomplete. There are assets that are licensed for example solely under CC0 for example.

Ok. I'll accept that.

> 
> But it's redundant either way (and the up to date license is available online). the full license info is generated in the textfield or accessible via the credits.csv. Moreover, the link to each license is now accessible via the license filters.

Good point.

> 
> Its plain overhead imo to put the same information in the readme, which you will also need to keep up to date.

Mostly you're correct. The only problem is that in theory some people might be taking assets directly from source files in the repo rather than the generator. Perhaps a link to the license filter is preferred?

> 
> > Also, can you add a statement that the source code in the repo is GPL 3?
> 
> This is already covered by the LICENSE file in the root of the repo

The license file is just a copy of GPL 3. It does not state what it applies to. i think we need to be clearer about that.

README.md
jrconway3: I think I would suggest adding something like:

"Please check credits.csv or the .json files for licensing info on each individual asset."

Something to get the effect across on how to find the licenses.

README.md
```diff
@@ -11,9 +11,9 @@ The Liberated Pixel Effort is a collaborative effort from a number of different
...
 
-The file `CREDITS.csv` lists the authors, license(s), and links to the original URL(s), for each image in `spritesheets`. **If you generate a sprite using this tool, you must credit all the authors**. You can do this one of two ways:
+The file `CREDITS.csv` lists the authors, license(s), and links to the original URL(s), for each image in `spritesheets`. **If you generate a sprite using this tool, you must credit all the authors (depending on the license)**. You can do this one of two ways:
```
jrconway3, line 16: Perhaps address what the ** means later.

```
** Assets licensed under CC0 do not require crediting the original artist, but it is still highly recommended to do so.
```

Something to that effect. I assume this is primarily about CC0.

README.md
jrconway3: Also maybe add ABOVE this line:

```
Universal LPC Spritesheet Character Generator source code is licensed under GPL 3.0. See LICENSE.txt.
```

jrconway3:
I agree with Gaurav. We need to clarify that it is the source code licensed under GPL 3.0 in the readme. I have one more comment to drop as well for the readme...

jrconway3:
Seems like I can't leave a new comment because its not in range, but this text:
> **Important: all art submitted to this project must be available under the [Creative Commons Attribution-ShareAlike 3.0](http://creativecommons.org/licenses/by-sa/3.0/) and/or the [GNU GPL 3.0](http://www.gnu.org/licenses/gpl-3.0.html) licenses**.

I feel like should be updated because assets do NOT have to be in CC-BY-SA 3.0 / GPL 3.0. Perhaps we should link to the OGA licensing page and say it must be one of the available licenses.

sanderfrenken:
> The license file is just a copy of GPL 3. It does not state what it applies to. i think we need to be clearer about that.

> Also maybe add ABOVE this line: 


@Gaurav0  @jrconway3 by having the license in the root, this is automatically picked up by GitHub web and app as a the repo's license. It's a standard way of putting license information in a (Github) repo.

See:

Scherm­afbeelding 2025-01-10 om 22 15 33

 and FYI:  https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/licensing-a-repository

sanderfrenken:
> Seems like I can't leave a new comment because its not in range, but this text:
> 
> > **Important: all art submitted to this project must be available under the [Creative Commons Attribution-ShareAlike 3.0](http://creativecommons.org/licenses/by-sa/3.0/) and/or the [GNU GPL 3.0](http://www.gnu.org/licenses/gpl-3.0.html) licenses**.
> 
> I feel like should be updated because assets do NOT have to be in CC-BY-SA 3.0 / GPL 3.0. Perhaps we should link to the OGA licensing page and say it must be one of the available licenses.

Yes I was also wondering what to do with this block. I think there aren't really license restrictions, as long as there is a license. Right?

README.md
sanderfrenken: I rather not go into such details here. Someone using these assets should themselves invest a little time in what a license implies when using a certain asset. Otherwise you will always give half or incomplete information here.

jrconway3:
No, there are licensing restrictions. All assets need to be in an open license. CC0, CC-BY, CC-BY-SA, OGA-BY, or GPL. Those are the supported licenses.

We can't support things like custom licenses for assets or assets that restriction commercial usage.



Thanks for the confirmation on the LICENSE.txt. If that's Github's standard and it shows the GPL 3.0 license by having it be in the index, then we're good.

README.md
jrconway3: Then we need to provide links to the original licenses somewhere that will clarify this. Perhaps provide a link to the OGA licensing page.

sanderfrenken:
> All assets need to be in an open license. CC0, CC-BY, CC-BY-SA, OGA-BY, or GPL. Those are the supported licenses.

> Then we need to provide links to the original licenses somewhere that will clarify this. Perhaps provide a link to the OGA licensing page.

> Mostly you're correct. The only problem is that in theory some people might be taking assets directly from source files in the repo rather than the generator. Perhaps a link to the license filter is preferred?


I will write a block then that names these licenses, including links to them, are the licenses we support for the assets in the generator. Moreover, I will emphasize that all artwork (under /spritesheets) in this repo is licensed under one or more of these licenses, and that the user should use the assets with respect to the license. 

I think that will cover all we need.

bluecarrot16:
I don't think any of these changes are particularly helpful (except that it's probably fine to remove the files in root containing the full text of the licenses). People are constantly wandering into the OGA forums (and my DMs) and asking very basic questions about the license. Making that information harder to find will do nobody any favors. 

If you're concerned about the statement "everything is available under CC-BY-SA 3.0 and/or GPL license(s)" isn't comprehensive, you can expand it; I might say "everything is available **at least** under CC-BY-SA 3.0 and/or GPL 3.0; some assets are available under other licenses too." which is the point this sentence is trying to make and is still true. You could also list the other licenses below and link to OGA for more information. Just saying "it's available under a license" is not helpful.

jrconway3:
"everything is available at least under CC-BY-SA 3.0 and/or GPL 3.0"

So in regards to this, its technically not true anymore. Some assets are available under CC0 which technically isn't under any other license. I do agree, though, that the updated changes aren't clear at all and simply saying "some licenses" is just confusing. We should list the available licenses and explain how to find them and the details of them.

I also agree that yes, users are constantly getting confused over the licenses and asking for clarifications. So making things more confusing is a problem.

sanderfrenken:
I pushed new changes, I think this is more complete than the information that was there. Have a look and let me know :)

Gaurav0 (approved):

jrconway3 (approved):
This looks good to me.

PR #61, opened 2025-01-12
Fix ordering in parse_zpos.js

sanderfrenken:
Ordering was depending on ordering of filesystem, which is fragile given it's OS influenced.


jrconway3 (dismissed):
I don't see anything innately wrong with it, but I won't know this works correctly until I can actually test it.

sanderfrenken:
you can cherry pick changes to your branch. You shouldn't see a diff in the zpos csv anymore then

jrconway3 (approved):
That fixed it, thanks a bunch!

PR #63, opened 2025-01-12
Slice source images to animations

sanderfrenken:


sources/chargen.js
```diff
@@ -514,13 +531,44 @@ $(document).ready(function() {
...
     }
     addCustomAnimationPreviews();
   }
 
+  function getDirectoryForAsset(filePath) {
+    const index = filePath.lastIndexOf("\/");
+    if (index > -1) {
+        return filePath.substring(0, index);
+    } else {
+      console.error("directory not ok");
```
Gaurav0, line 556: Let's print out the directory here.

sources/chargen.js
```diff
@@ -514,13 +531,44 @@ $(document).ready(function() {
...
+      return undefined;
+    }
+  }
+
+  function getFileNameForAssetPath(filePath) {
+    const index = filePath.lastIndexOf("\/");
```
Gaurav0, line 562: '/'

sources/chargen.js
```diff
@@ -514,13 +531,44 @@ $(document).ready(function() {
...
+    if (index > -1) {
+        return filePath.substring(index);
+    } else {
+      console.error("filename not ok");
```
Gaurav0, line 566: print out the filename here.

scripts/imageProcessing/sliceImages.js
```diff
@@ -0,0 +1,104 @@
+const fs = require('fs');
+const { execSync } = require('child_process');
+
+const universalFrameSize = 64;
+const universalSheetWidth = 832;
+
+function cropToSubSheet(sheet, width, height, offsetX, offsetY, targetDir) {
+    const index = sheet.lastIndexOf("\/");
```
Gaurav0, line 8: if you use single quotes, you won't have to escape the slash.

'/'

scripts/imageProcessing/sliceImages.js
```diff
@@ -0,0 +1,104 @@
...
+    var directory = "";
+    if (index > -1) {
+        directory = sheet.substring(0, index);
+    }
```
Gaurav0, line 12: This lastIndexOf / substring code is repeated below. Let's extract if to a function.

scripts/imageProcessing/sliceImages.js
```diff
@@ -0,0 +1,104 @@
...
+    const newFile = `${directory}/${targetDir}${sheet.substring(index)}`;
+    execSync(`magick ${sheet} -crop ${width}x${height}+${offsetX}+${offsetY} ${newFile}`, (err) => {
+        if (err) {
+            throw err;
+        }
+    });
+}
+
+function cropSheet(sheet) {
+    cropToSubSheet(sheet, 7*universalFrameSize, 4*universalFrameSize,0,0,"spellcast");
```
Gaurav0, line 25: Let's refactor this. the only things that change per line are the number of frames and the name of the sheet.

So create a map:

```
const animToFramesMap = {
  "spellcast": 7,
  "thrust: 8,
  ...
}
```

Then:

```
Object.entries(animToFramesMap).forEach(([anim, frames] =>
  cropToSubSheet(sheet, frames*universalFrameSize, 0, 4,*universalFramesSize, anim));
```

sources/chargen.js
```diff
@@ -514,13 +531,44 @@ $(document).ready(function() {
         }
         ctx.drawImage(img, 0, universalSheetHeight+offSetInAdditionToOtherCustomActions);
       } else {
-        drawImage(ctx, img);
+        console.log(filePath);
```
Gaurav0, line 534: Once you're finished debugging, don't forget to remove these.

scripts/imageProcessing/sliceImages.js
Gaurav0: Actually, you do something like this already in chargen.js

scripts/imageProcessing/sliceImages.js
sanderfrenken: yeah this script will be a one-timer, I will remove it before publishing this PR. So no need to look at it from CQ perspective. As long as it gets the job done ;)

scripts/imageProcessing/sliceImages.js
sanderfrenken: > Let's refactor this. the only things that change per line are the number of frames and the name of the sheet.

That is not true here, also the height (amount of animation rows), offSetX and offsetY is different per animation.

Gaurav0:
The diff is very difficult for github to load. Could you put the script changes in one commit, and the results of running it in another?

scripts/imageProcessing/sliceImages.js
Gaurav0: Fair enough.

sanderfrenken:
> The diff is very difficult for github to load. Could you put the script changes in one commit, and the results of running it in another?

I am creating a new PR. Indeed it will separate the scripts in 1 commit and the run in another. It takes 4 hours now to run the script..

sanderfrenken:
I will close this PR later when the new one is ready

sanderfrenken:
new PR: https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/pull/71

PR #64, opened 2025-01-13
Fix item selection

sanderfrenken:
Currently when you select anything, the body type switches back to "Male". When you then select "Female" for instance, it is not processed. Only after a second click it does process.

I don't know why this happens exactly, but I tracked it down to this change that was recently made. Reverting this fixes the issue.

### DIFF
--- a/sources/chargen.js
+++ b/sources/chargen.js
@@ -5,8 +5,6 @@ $.expr[':'].icontains = function(a, i, m) {
       .indexOf(m[3].toUpperCase()) >= 0;
 };
 
-$.fn.reverse = [].reverse;
-
 $(document).ready(function() {
 
   var matchBodyColor = true;
@@ -584,7 +582,7 @@ $(document).ready(function() {
   }
 
   function interpretParams() {
-    $("input[type=radio]").reverse().each(function() {
+    $("input[type=radio]").each(function() {
       var words = _.words($(this).attr('id'), '-');
       var initial = _.initial(words).join('-');
       $(this).prop("checked", $(this).attr("checked") || params[initial] == _.last(words));


sanderfrenken:
See:

https://github.com/user-attachments/assets/35a8592b-c578-44c1-840e-ee5917d9e52b

Gaurav0 (commented):
Yeah, I made this change. The problem I was trying to solve was that if multiple radio buttons had the same id, the last one was selected. For example, set no body and refresh and the radio button named body-none next to the skeleton was the one that ended up selected, rather than the body-none next to the male body.

Let's see if we can solve the problem another way.

jrconway3:
> Yeah, I made this change. The problem I was trying to solve was that if multiple radio buttons had the same id, the last one was selected. For example, set no body and refresh and the radio button named body-none next to the skeleton was the one that ended up selected, rather than the body-none next to the male body.
> 
> Let's see if we can solve the problem another way.

To my knowledge, there aren't supposed to be any duplicate ID's on the same webpage. The purpose of `id` as an attribute is that it is a 100% unique attribute on the entire web page. So if that's the issue you're trying to solve, we should be solving the problem of their being a duplicate id to begin with.

jrconway3:
How long does it take for it to take effect on github.io? If its immediate, then we might want to add a version number to chargen.js to force it to reload. Anyone with that file cached is going to still have the same issue.

Gaurav0:
> 
> To my knowledge, there aren't supposed to be any duplicate ID's on the same webpage. The purpose of `id` as an attribute is that it is a 100% unique attribute on the entire web page. So if that's the issue you're trying to solve, we should be solving the problem of their being a duplicate id to begin with.

There never used to be. This mostly affects -none ids.

PR #67, opened 2025-01-13
alternate fix for jumping body-none

Gaurav0:
Description of bug:

Select "No body" under body color. Hit Refresh. "Expand Selected" Expected result: "No body" next to "Body Color" is selected. Actual result: "No body" next to "Skeleton" is selected.

A better fix would be to ensure that no radio buttons have the same id.

This at least makes sure that the first such id gets selected instead of all of them in order, ending with the last.

### DIFF
--- a/sources/chargen.js
+++ b/sources/chargen.js
@@ -582,10 +582,15 @@ $(document).ready(function() {
   }
 
   function interpretParams() {
+    const ids = new Set();
     $("input[type=radio]").each(function() {
-      var words = _.words($(this).attr('id'), '-');
-      var initial = _.initial(words).join('-');
-      $(this).prop("checked", $(this).attr("checked") || params[initial] == _.last(words));
+      const id = $(this).attr('id');
+      if (!ids.has(id)) {
+        ids.add(id);
+        const words = _.words($(this).attr('id'), '-');
+        const initial = _.initial(words).join('-');
+        $(this).prop("checked", $(this).attr("checked") || params[initial] === _.last(words));
+      }
     });
   }
 


jrconway3:
Thanks for explaining the issue. I'll run ahead and test this before and after pulling the branch.

jrconway3 (changes requested):
I pulled the update and tried it out, and I'm still getting the same issue? Its still selecting the skeleton "none" instead of the base body.

sanderfrenken:
So this issue only occurs with the none-* variants, and I doubt it's truly an issue at all.
You report this issue for the body selection, but you can reproduce it with a hair style as well.

The thing is, there are multiple asset groups that manipulate the same "asset type", which in the json files we define as `type_name`. For instance, all hair styles have the `type_name` *hair*, all body types have the `type_name` *body*. 

We have the default group of body variants, with the standard colors. But we also have two special types, zombie and skeleton.

Assets that share the same `type_name`, have a shared state. For instance, you can not apply two hairstyles at the same time, and you can also not be a green body color variant and a special zombie variant at the same time.

As such, clicking `no body` will reset the body type_name to none, removing any selected body type. Its essential that the id matches When interpret params runs, it indeed checks for the first hit, and that can be the wrong one. 

The logic works in such a way that the id of the none checkbox contains the type_name in order to toggle off that type_name regardless of which variant is selected.

A better solution would be to assign unique Id's, but I need to dig in the code how it works again. 

Alternatively, functionally it would also be correct to actually select all none-type_name id's present in the dom for a given type_name when none is selected.

sanderfrenken:
@Gaurav0 @jrconway3  please check this PR as a solution for the non-unique ID's 

https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/pull/69

sanderfrenken (changes requested):
Please check my open PR that assigns unique ID's. I think that is more resilient.

PR #71, opened 2025-01-14
Slice images def run

sanderfrenken:


sanderfrenken:
In this PR all images are sliced to single animation files. It contains > 300K changes, and as a result at least on my browser, I can not check the second commit or the changed files.

You can check out this branch though and run locally. Managing this branch has become a bit painful as pushing changes takes forever. So I hope to merge it asap and for the meantime, don't merge other branches (sorry for that, but I really dont want to mess with this branch and risk that I need to rerun the whole script. It took > 5 hours to complete).

In the first commit is the most logic. The processImage script I suspect/ hope we wont need anymore, but I keep it for now so we can always change/ use it later if needed again. When this is merged, I will add a readme next to the script.

Inside charges.js I made some changes, ofc to load the images now separately, and to fix the previews. It seems to work. At least, on my machine ;) 

Anyways, if you find anything, I would propose to consider to merge as is if possible, and then fix forward. For the same reason as mentioned earlier, this branch is too large already.

jrconway3:
Sounds good. I'll pull it and run some tests. Perhaps what we should do is if there's still more issues, we open cards to address those later.

jrconway3:
Honestly it looks really good. One asset that's missing is child pants, but its literally just walk only anyway. Not sure why these didn't split.

sanderfrenken:
> Honestly it looks really good. One asset that's missing is child pants, but its literally just walk only anyway. Not sure why these didn't split.

that is good to hear :) 

I see why. Those sheets have an inconsistent width: they are 768 pixels width.
The image processing script has skipped it because of that. 

Do you think you can fix the child pants later?

bluecarrot16:
I don't have time to look at this right away, are you sure performance will be better once it's merged? Will this just introduce a huge change into `main` that will make the already huge repo slower to clone/manage?

I'm also interested to know about performance of loading/running the generator too, since we've had issues with that in the past. 

Let's take a breath here and also make sure we identify as many issues as possible before merging so that we don't have assets silently disappear and discover them months later.

jrconway3:
Yeah that's not hard. The child pants are only walk animation anyway, so I can easily just crop it down to size and add repalettes again with lpctools.

jrconway3:
> I don't have time to look at this right away, are you sure performance will be better once it's merged? Will this just introduce a huge change into `main` that will make the already huge repo slower to clone/manage?
> 
> I'm also interested to know about performance of loading/running the generator too, since we've had issues with that in the past.
> 
> Let's take a breath here and also make sure we identify as many issues as possible before merging so that we don't have assets silently disappear and discover them months later.

I do have concerns about this as well. That being said, I ran tests locally and the performance seems fine, it doesn't seem any slower than it was before. I was concerned that splitting the assets into different layers might cause performance issues, but personally speaking I am not seeing that.


However, I definitely do think it'd be ideal to get as many people looking into it as reasonably possible to make sure everything looks good before merging.

sanderfrenken:
> I don't have time to look at this right away, are you sure performance will be better once it's merged? Will this just introduce a huge change into main that will make the already huge repo slower to clone/manage?

The size of the repo did not change, so I don't expect any impact on cloning/ managing once merged. But my tooling has a hard time now managing the large diff.

> I'm also interested to know about performance of loading/running the generator too, since we've had issues with that in the past.

As far as I can see the performance is not notably altered. Did you see notice significant negative impact @jrconway3 ?

Maybe you can actually somehow branch this to one of your forks, and then check the performance via your GitHub pages? I only tested this locally.

> Let's take a breath here and also make sure we identify as many issues as possible before merging so that we don't have assets silently disappear and discover them months later.

Yeah sure, please have a look if/ when you can. I just prefer not to merge conflicting items in between in order to prevent having to rebase this branch.

jrconway3:
The thing is... there is something called "version history" that's a thing. Its something I've been concerned with as well. If you make huge sweeping changes to pre-existing assets, github's version history will still keep the old stuff present, as well. I would assume version history means that any old versions are still part of the repository size. That being said, I can't see anything that points to the size of the repo in an obvious location.


I can try deploying my fork to github pages to test I suppose.

sanderfrenken:
> I can try deploying my fork to github pages to test I suppose.

That would be really helpful

> version history

Yes that is true, but we could eventually prune the history IMO if this really becomes an issue.

sanderfrenken:
I found sth I didn't know about, you can deploy to github pages also a specific branch. I just tried and it works well. I will now switch it back to main though. If you want, I can deploy this branch for a while so we can test.

sources/chargen.js
```diff
@@ -684,20 +725,27 @@ $(document).ready(function() {
         const previewToDraw = {};
         const animation =  $(this).data(`layer_1_custom_animation`);
 
-        if($(this).data(`layer_1_${getBodyTypeName()}`) === undefined){
-          previewToDraw.link = $(this).data(`layer_1_${getBodyTypeName()}`);
+        if($(this).data(`layer_1_${getBodyTypeName()}`) === undefined) {
+          var imageLink = $(this).data(`layer_1_${getBodyTypeName()}`);
```
Gaurav0, line 729: Let's (pun intended) avoid var in new code. Use let.

sources/chargen.js
```diff
@@ -684,20 +725,27 @@ $(document).ready(function() {
...
+          if (imageLink !== undefined) {
+            imageLink = updatePreviewLink(imageLink);
```
Gaurav0, line 731: These lines can be: `imageLink = imageLink && updatePreviewLink(imageLink);`

sources/chargen.js
```diff
@@ -512,7 +533,27 @@ $(document).ready(function() {
...
+        const assetName = getFileNameForAssetPath(filePath);
+
+        for (const [key, value] of Object.entries(base_animations)) {
+          var animationToCheck = key;
+          if (key === "combat_idle") {
+            animationToCheck = "combat";
+          } else if (key === "backslash") {
+            animationToCheck = "1h_slash";
+          } else if (key === "halfslash") {
+            animationToCheck = "1h_halfslash";
```
Gaurav0, line 546: Can we use a map instead of several if/else statements?

Gaurav0:
Btw, feel free to merge and address my review comments later.

jrconway3:
> I found sth I didn't know about, you can deploy to github pages also a specific branch. I just tried and it works well. I will now switch it back to main though. If you want, I can deploy this branch for a while so we can test.

https://jrconway3.github.io/Universal-LPC-Spritesheet-Character-Generator/#?body=Body_color_light&head=Human_male_light

sources/chargen.js
sanderfrenken: yes well ideally I would rename 
`1h_slash` to `backslash`
`1h_halfslash` to `halfslash` 
`combat` to `combat_idle`

And then I can remove this ugly if-else. I will do that once we merged this one ok :)

sanderfrenken:
> Btw, feel free to merge and address my review comments later.

I will address a couple already, I had to push a fix anyways!

sources/chargen.js
sanderfrenken: Fixed it (see: https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/pull/71/commits/a4459de5ab849becfecb87bb67d92423516a3872)

sanderfrenken:
I found that when using oversized weapons, the drawing was slower than usual. Digging into it, I found the root cause and pushed fixes in this commit:

https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/pull/71/commits/325e8f6dc7639260dcf148f5be6deb6714f482ad

The logic to resolve the issues would already be of benefit for the current implementation as well. This  is a bug that was already present, but has become more visible with the new setup.

So the problem is that when we draw() the items, some of the images need to be loaded (we also keep an in memory cache of previously loaded assets). When an image is loaded, it will invoke the method to redraw() everything again. This is very costly: If there are 4 new assets to be loaded, redraw() is invoked 4 times.

 Especially with custom animations this is bad, as they are drawn to the canvas on calculated positions. Now with the new setup, which uses cut-up sheets, the 4 new assets become 4*8( example, it can be more or less depending on the supported animationTypes) = 32 times a redraw() invocation.

To prevent this, I rewrote the logic. First, when all selected assets are processed, the generator will load them into memory. On load of an image, a function is invoked that will check if all images are loaded. if that is the case, it will invoke the draw() method actually once.

I added console.logs here and there where appropriate, so you can assess the functioning. I think it's an important improvement. As far as I can see now the performance is much better, possibly even better than before using the non-sliced images (though ofc if we bring this change to the main, it will also become more performant without the sliced images).

@jrconway3  would it be possible for you to deploy this version to your fork? I also fixed the previews and most of the comments of @Gaurav0

jrconway3:
I deployed. Oversized animations seem to load much faster now, that being said, Arming Sword appears to be using the wrong animations in Backslash (and Halfslash is missing). I suspect its only happening for these two, the Arming Sword looks fine on Slash.

sanderfrenken:
> I deployed. Oversized animations seem to load much faster now, that being said, Arming Sword appears to be using the wrong animations in Backslash (and Halfslash is missing). I suspect its only happening for these two, the Arming Sword looks fine on Slash.

ah let me check

Gaurav0:
Intermittent issue, can't reproduce reliably:

Screenshot 2025-01-15 at 10 12 32 AM

sanderfrenken:
> Intermittent issue, can't reproduce reliably:
> 
> Screenshot 2025-01-15 at 10 12 32 AM

This https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/pull/71/commits/ebcedb86761705c45be382c912e1c14c266a1aad should fix it. But yes. It's not nice code wise. But yes. It does work. 

Hmmm 😩

jrconway3:
I think it might have been the issue Gaurav reported. I went back and tested again and backslash Arming Sword worked, but the head was missing on half the animations. Then the head was back again. I pulled and pushed this update again, so hopefully its live shortly.

Gaurav0:
I have a fix incoming.

Gaurav0:
Files changed = ∞ lol

Gaurav0:
https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/pull/74

sanderfrenken:
> #74

Hmm I merged it but it messes up:

https://github.com/user-attachments/assets/a9d9b739-4284-4d73-b32a-79370055e936

Its calling it self recursively now.

sanderfrenken:
got it:

https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/pull/71/commits/af5583ec10a233ccdaad910bd6c1a51ca91befd7

@jrconway3 can you deploy this version when you have time? Sorry to bother again. I think we are almost there....

jrconway3:
Looks like I just did by mistake. I pulled and pushed the updates because I thought I was pulling/pushing the merge, but I guess I caught your latest push at the same time.

jrconway3:
Ah I see what's happening. I reported on @Gaurav0 's PR that when you click on an oversize animation its stuck there. However, I can switch between the oversized animations. But I can't go back to a non-oversized animation:
https://jrconway3.github.io/Universal-LPC-Spritesheet-Character-Generator/#?body=Body_color_light&head=Human_male_light&weapon=Arming_Sword_ceramic

sanderfrenken:
> Ah I see what's happening. I reported on @Gaurav0 's PR that when you click on an oversize animation its stuck there. However, I can switch between the oversized animations. But I can't go back to a non-oversized animation: https://jrconway3.github.io/Universal-LPC-Spritesheet-Character-Generator/#?body=Body_color_light&head=Human_male_light&weapon=Arming_Sword_ceramic

Isnt that the same already on master? The previews are a bit flaky (needs to be fixed one day)

jrconway3:
Nvm, I guess it is on master. I never noticed it before. So that wasn't caused by Gaurav, but I'll open a card for that issue as a separate thing.

sanderfrenken:
Well I can't find any remaining issues, but take your time to look properly :)

jrconway3:
I'll test all options tonight and also open a PR fix for the assets that didn't split properly (like those child pants). I doubt the PR will be that big, so we should be good there. I might want to remove the assets that generated blank sheets, too, but that could wait. If the PR is too big when including "blank" asset removals, then I won't open that PR until later.

sanderfrenken:
@jrconway3 this PR now also includes the checkboxes for animations filter, can you deploy this version when you have time?

sanderfrenken:
@Gaurav0  also made a fix for the preview animations, that is now in here as well.

Gaurav0:
I deployed it here: https://gaurav0.github.io/Universal-LPC-Spritesheet-Character-Generator/#?body=Body_color_light&head=Human_male_light

sanderfrenken:
looks great! Thanks @Gaurav0

jrconway3:
Sorry guys, I got distracted yesterday. Anyway, I deployed the latest changes to mine now.

Since my branch already got merged I can probably start working on clearing out the blank assets. I also noticed an error when looking at the fairy wings, but this is pre-existing. There's a color palette that doesn't exist on many of them. I was trying to go through every asset and make sure everything works before we merge.

sanderfrenken:
I think the generator will be in a much better shape after all this work is done :)

jrconway3:
Splitting the full sheets into smaller sheets is definitely going to be a pretty big benefit. That will make it easier to add more "partial" sets of assets and make it look less like "omg why are half the assets missing?!" because instead the inconsistent assets will be more interspersed.

I'll probably drop a big update soon after this gets merged (within a week or two) filling out a lot of stuff in idle and at least the first two frames of emotes.

jrconway3:
Yup, preview animations look great here:
https://jrconway3.github.io/Universal-LPC-Spritesheet-Character-Generator/#?body=Body_color_fur_copper&head=Wartotaur_fur_copper&weapon=Arming_Sword_ceramic&shadow=Shadow_shadow&wound_brain=none_Brain&wound_ribs=none_Ribs&wound_eye=none_Eye&wound_mouth=none_Mouth&prosthesis_leg=none_Peg+leg&prosthesis_hand=none_Hook+hand&tail=none_Lizard+tail&wings=none_Batlike+Lizard+Wings&wings_edge=none_Monarch+Wings+Edge&wings_dots=none_Monarch+Wings+Dots&wound_arm=none_Arm

You can now swap out of "custom" animations just fine.

jrconway3:
Interesting. Apparently if the animation isn't in the animation list, then it isn't drawn. However, if it is included, that does mean it will throw an error if it can't find it. I had to fix zombie, skeleton, and jack heads because they didn't have the full sheet of animations.


This also causes problems if I remove the blank sheet, though... well, in theory. It throws errors, but I don't know if that's really a big deal. However, perhaps I shouldn't remove blank sheets because of this. This is primarily an issue with the body sheets, or where certain sheets aren't fully meshed between all body types.

jrconway3:
Pushed some more updates. I removed the blank sheets I found so far, mostly in the "bodies" section. I also removed shoot and emotes from muscular because I didn't actually do anything there yet; I only set the idle frame for shoot and a basic stand pose for emotes, so there was no point in keeping these. They weren't accurate anyway.


For "child" I also fixed the child shirts cropping, and I restricted the animations list for child shirt, child pants, and child skirt to only "walk" (because that's all that's there) so they don't show errors on other animations. Only the body itself shows those errors.


I'll keep going and examining other assets. I reviewed everything up TO hairstyles (so everything ABOVE hair). Then I went through a good chunk of the hairstyles already and haven't seen any issues so far.

I reviewed all hairstyles in afro, curly, bald/shaved, and short. I haven't reviewed Hair Extensions or the remaining hairstyles below short yet, nor any assets after hairstyles (except for child assets).

Gaurav0:
According to https://www.imagemagick.org/discourse-server/viewtopic.php?t=33624 you can detect transparent blank images using `image-magick convert image -channel a -separate -scale 1x1! -format "%[fx:mean]\n" info::`

jrconway3:
All Hairstyles and Appendages look good, too. Hats up next.

sanderfrenken:
Once this is merged, I might also add a script to the repo that one could use to combine all images back to universal sheets. I at least need that for my own games, so I might just as well contribute it here. But it will have to wait for a bit, so I wont add it to this PR directly

jrconway3:
Pushed the latest updates to my fork.

jrconway3:
Pushed the latest updates.

sanderfrenken:
nice improvement on the scrolling @Gaurav0

Gaurav0:
I have a script working to delete transparent blank pngs. PR incoming.

jrconway3:
I think I already got most of them anyway, but there may be a handful in, like, Muscular missing "shoot" shirts and pants because those frames don't exist on Muscular.

Gaurav0:
> I think I already got most of them anyway, but there may be a handful in, like, Muscular missing "shoot" shirts and pants because those frames don't exist on Muscular.

You missed about 25k of them.

jrconway3:
Seems like most of this are "climb" (of which many assets don't have "climb" because they only appear in the front, such as bears or facial assets) and weapons because weapons are typically only on walk and the respective attack animation.

I didn't even think to check for climb. And as for weapons, I hadn't gotten to weapons yet, but even then, I wasn't even thinking of checking the sheets in weapons.

sanderfrenken:
> Seems like most of this are "climb" (of which many assets don't have "climb" because they only appear in the front, such as bears or facial assets) and weapons because weapons are typically only on walk and the respective attack animation.
> 
> 
> 
> I didn't even think to check for climb. And as for weapons, I hadn't gotten to weapons yet, but even then, I wasn't even thinking of checking the sheets in weapons.

Hmm so if i understand it correctly, these are "functionally correct" empty images. The assets are supporting those animations, but there is no pixel to be drawn for that animation. Removing those images, wouldn't that be incorrect?
The generator will still try to load them as the animation is supported for that asset.

Gaurav0:
> Hmm so if i understand it correctly, these are "functionally correct" empty images. The assets are supporting those animations, but there is no pixel to be drawn for that animation. Removing those images, wouldn't that be incorrect? The generator will still try to load them as the animation is supported for that asset.

It honestly never occurred to me that an animation could be supported and legitimately be blank.

sanderfrenken:
> It honestly never occurred to me that an animation could be supported and legitimately be blank.

No I understand, also didn't think about it. Not sure how to proceed with it. You could argue, that eg the climb animation for a beard is legitimately missing, and therefore a blank image is intended. For weapons however, you could argue that for instance a spell cast animation that is fully blank for a dagger, is not legitimate. The spell cast animation could in theory just as well be made with a dagger in hand as well..

Curious to hear your thoughts on this.

jrconway3:
Sorry, guys, I should've commented on this last night.

In any case, I see no reason to "keep" any blank sheets. However, it could be ideal to implement a check somehow in the sheet definitions to say an animation is supported, but not to search for a sheet.

jrconway3:
Btw, I merged Gaurav's updates to my fork last night, and I merged the latest update here now, so both should be live and we can test those updates.

jrconway3:
I have concerns with the idea of removing assets from supporting certain animations, but perhaps it still could work. Removing "Climb" from assets that don't have anything, even if it still works, could be fine as Supported Animation should still allow Climb to appear anyway. My only concern would be the reverse--removing things like Glasses because Climb doesn't exist on Glasses.

jrconway3:
Halfmoon Glasses are missing on Down, but that's my fault. Apparently I never added them to the original sheet.


For assets that are things I want to fix later or were things I did, but messed something up, I'm opening issues over here:
https://github.com/jrconway3/Universal-LPC-Spritesheet-Character-Generator/issues

So that way I can keep track of them.


If there's an issue caused by this PR I want to fix it before merging, though.

Gaurav0:
Oversize weapons previews in the left column don't work in this PR.

jrconway3:
I'm not sure what you're talking about? What's this about the "left column"? I don't see how oversize weapons are relevant to the left column at all.

Gaurav0:
Screenshot 2025-01-18 at 1 45 37 PM

jrconway3:
Yeah I noticed it happening on Scimitar. That being said... I suspect this might be happening on the original, too.

https://liberatedpixelcup.github.io/Universal-LPC-Spritesheet-Character-Generator/#?body=Body_color_light&head=Human_male_light&hair=Side_Swoop_sandy

Yup, it is. Scimitar, Katana, and Saber are not showing up there. Rapier is, however. Its because these don't have walking or down animations which is what the ones that had a "universal" sheet do have.


So its not relevant to this branch, but the entire editor as a whole.

jrconway3:
It seems I stand corrected. The polearms work on master, but scimitar/katana/longsword do not.

sanderfrenken:
> I'm not sure what you're talking about? What's this about the "left column"? I don't see how oversize weapons are relevant to the left column at all.
> 
> EDIT: Okay, scimitar is the first one I found. There's no preview because apparently its ONLY in oversized?

Yeah there is a specific logic for that in `chargen.js`, but apparently that partly broke (it didn't work for all assets anyways before either). That can be fixed later, it's not urgent IMO.

sanderfrenken:
> I have concerns with the idea of removing assets from supporting certain animations, but perhaps it still could work. Removing "Climb" from assets that don't have anything, even if it still works, could be fine as Supported Animation should still allow Climb to appear anyway. My only concern would be the reverse--removing things like Glasses because Climb doesn't exist on Glasses.

I think it will just become confusing if we start removing animations, even though the png is blank. If they are supported, there should be an image, that is my opinion. 

If you do start to make exceptions to that, then you will need to add extra information into the sheetDefinition. Like an ignore list for certain animations or whatever. Yes it can all be done. But I think it will just be confusing.

sanderfrenken:
> > I'm not sure what you're talking about? What's this about the "left column"? I don't see how oversize weapons are relevant to the left column at all.
> > EDIT: Okay, scimitar is the first one I found. There's no preview because apparently its ONLY in oversized?
> 
> Yeah there is a specific logic for that in `chargen.js`, but apparently that partly broke (it didn't work for all assets anyways before either). That can be fixed later, it's not urgent IMO.

This is the method:

`drawPreviews()`

And it's a mess. Should be refactored. But that can be done later, its not relevant to this PR which is already out of its bounds.

jrconway3:
I don't like the idea of having a blank image just to support a certain frame. The blank images are just a waste of space and are unnecessary. As long as the error doesn't slow down the editor I don't see a problem with leaving the errors there. Its not like anybody is going to see the error unless you actively go out of your way to look in the console which most end-users won't be doing.


We can use the errors to check for certain things, but if its something that should be missing (such as climb on assets without a north-facing asset) then I say let the error be thrown. I'd rather not be including blank assets for no reason.

sanderfrenken:
> We can use the errors to check for certain things, but if its something that should be missing (such as climb on assets without a north-facing asset) then I say let the error be thrown.

That will just make it very hard to find real errors.. Maybe an end user will not see it but as a developer I really dont advocate throwing errors all over the place that are functionally no error at all.

Moreover, an end user that will not use the generator but looks up the images in the sprite sheets directory will be confused to see certain files missing. 

I get the point you want to get rid of empty images, but I also see significant downsides for doing so.

Gaurav0:
> That will just make it very hard to find real errors.. Maybe an end user will not see it but as a developer I really dont advocate throwing errors all over the place that are functionally no error at all.
> 
> Moreover, an end user that will not use the generator but looks up the images in the sprite sheets directory will be confused to see certain files missing.
> 
> I get the point you want to get rid of empty images, but I also see significant downsides for doing so.

We can use a debug flag set to true when running on localhost and false otherwise. When enabled, it throws the error, when not, it silently ignores it.

`const DEBUG = window.location.hostname === 'localhost';`

```
try {

} catch (e) {
  if ((pseudocode: e is FileNotFound) && DEBUG) throw e;
}
```

Eventually, let's get rid of the empty images and make it so they don't throw errors, but let's not let it hold up the PR.

jrconway3:
Removing blank images isn't holding up the PR for me. I'm just trying to make sure we follow bluecarrot16's advice and make sure nothing is MISSING.

My only concern with the blank images PR is if it accidentally removed legitimate assets, but I think its fine. I just approved that PR.

jrconway3:
> Eventually, let's get rid of the empty images and make it so they don't throw errors, but let's not let it hold up the PR.

I agree with this as well. Whether we remove empty images now or not, I don't think it matters. And I don't think it matters if we temporarily allow errors for things like "climb" either. We can fix those issues later without holding up this PR.

I think its better to remove "climb" and fix the errors later. But I also agree I think its better to suppress the errors on master anyway, and only allow errors on local.


In any case...

I'm gotten through Hats now, I'll look at Helmets next. I just want to do a quick run through of the remaining assets then approve this PR. I think we should be good here, though. Its the weekend so I have more free time to finish this review.

sanderfrenken:
> > That will just make it very hard to find real errors.. Maybe an end user will not see it but as a developer I really dont advocate throwing errors all over the place that are functionally no error at all.
> > Moreover, an end user that will not use the generator but looks up the images in the sprite sheets directory will be confused to see certain files missing.
> > I get the point you want to get rid of empty images, but I also see significant downsides for doing so.
> 
> We can use a debug flag set to true when running on localhost and false otherwise. When enabled, it throws the error, when not, it silently ignores it.
> 
> `const DEBUG = window.location.hostname === 'localhost';`
> 
> ```
> try {
> 
> } catch (e) {
>   if ((pseudocode: e is FileNotFound) && DEBUG) throw e;
> }
> ```
> 
> Eventually, let's get rid of the empty images and make it so they don't throw errors, but let's not let it hold up the PR.

if you could make sth like that work, that would be nice. I tried a while back to catch file not found exceptions, but I couldn't prevent the browser from logging these console.errors.

Anyways, let's go ahead then with removing the empty assets. If you both think its better then I respect that ofc.

jrconway3:
Okay I think I got everything now. I reviewed primarily Male and Female. Almost all "Teen" assets are identical to Female anyway. I didn't really check much in Muscular or Pregnant. I only checked head assets (hats, heads, ears, noses, hair, etc) once over.


If there is anything else broken, its going to be really hard to find it. I opened a new issue listing off all broken previews as well, we can address that issue later.

sanderfrenken:
Lovely :) So we are good to merge id say?

jrconway3:
If I can figure out how to approve it lol...

sanderfrenken:
> If I can figure out how to approve it lol...

hahahaha indeed. We can merge without approvals. @Gaurav0 you also ok right?

Gaurav0:
I approve. Let's ship it.

bluecarrot16:
sorry for insisting on holding this back but then never giving a formal "review." I'm glad we left a little more time because it seems a number of issues were identified and fixed. nice work all!

PR #74, opened 2025-01-15
fix image load timing issue

Gaurav0:
While images are being loaded, a request to load images cannot "enter the queue". This is a fix for that.

### DIFF
--- a/sources/chargen.js
+++ b/sources/chargen.js
@@ -13,13 +13,13 @@ $(document).ready(function() {
   var params = jHash.val();
   var sheetCredits = [];
 
-  var imagesToLoad = 0;
-  var imagesLoaded = 0;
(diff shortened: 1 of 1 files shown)


jrconway3:
I deployed this fix here to test it:
https://jrconway3.github.io/Universal-LPC-Spritesheet-Character-Generator/#?body=Body_color_light&head=Human_male_light&weapon=Arming_Sword_ceramic


I'm not seeing random loading issues anymore, however, if I choose an "oversized" animation and switch to a non-oversized animation, it stays on the oversized animation. This only happens if you chose an oversized animation such as:
slash_128
backslash_128
halfslash_128


All three of these get locked into place if you choose the oversized animation on the link above. Other than this everything works fine.

Issue #79, opened 2025-01-17
Scrolling can be unwieldy


Gaurav0:
It would be nice to be able to scroll left and right columns independently, or to keep top in view.

sanderfrenken:
yeah couldn't agree more. Would indeed be nice to have the action buttons/ checkboxes on the top left floating, while the remainder of the left column could scroll independently from the canvas

jrconway3:
This is live now, but how is everyone feeling about it? Do we want to implement improvements or anything?


Having different scrolls is nice, but I do wonder if there's still ways to improve it... that being said, it would be difficult to improve it on different screen sizes.

Gaurav0:
I definitely think we can do better with smaller screen sizes. But I think we should open a new issue.

PR #81, opened 2025-01-17
Scrolling improvements

Gaurav0:
For #79 

- Sticky positioning on header
- Independent scrolling of columns

Preview:

Screenshot 2025-01-17 at 3 34 15 PM

### DIFF
--- a/sources/chargen.css
+++ b/sources/chargen.css
@@ -26,7 +26,22 @@ body {
 }
 
 header {
-  width: 640px;
+  position: sticky;
+  top: 0;
+  left: 0;
+  width: 100%;
+  display: flex;
+  flex-direction: row;
+  background-color: white;
+  opacity: 1;
+  z-index: 4;
+}
+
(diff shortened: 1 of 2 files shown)


jrconway3:
In practice I don't like this very much. I'd prefer having much more room on the spritesheet itself. Right now the spritesheet is so low compared to the preview animation with a HUGE blank gap there. It doesn't look good and I'd much prefer the spritesheet going all the way to the top (right below the Preview Animation where it used to be).


My preference would be a 50/50 split with the header, buttons, and links being on the left, and the right being the Preview + Spritesheet. The Footer area is fine.

One possibility, though:
Split "Import / Export" into two separate rows, so that way we can keep the left column as narrow as possible. If Import / Export were two separate options on separate lines, then the scrollbar could go straight up to the top and still keep the header area in the left column.

Gaurav0:
@jrconway3 I'll see if I can do that.

PR #83, opened 2025-01-18
Remove blank transparent pngs using script

Gaurav0:


jrconway3 (approved):
I think this looks good. I'm not seeing any major issues with this.

jrconway3:
That being said, I think its better for Sanderfrenken to approve first before we merge. However, my branch does have it merged already, so that could be a potential hangup.

PR #87, opened 2025-01-19
debugging improvements

Gaurav0:
Also documented in code:

DEBUG mode will be turned on if on localhost and off in production
but this can be overridden by adding debug=(true|false) to the querystring.

| debug | isLocalhost | result |
| ------ | ----------- | ------ |
| true | true | true |
| true | false | true |
| false | true | false |
| false | false | false |
| unset | true | true |
| unset | false | false |

### DIFF
--- a/sources/chargen.js
+++ b/sources/chargen.js
@@ -5,6 +5,26 @@ $.expr[':'].icontains = function(a, i, m) {
       .indexOf(m[3].toUpperCase()) >= 0;
 };
 
+// DEBUG mode will be turned on if on localhost and off in production
+// but this can be overridden by adding debug=(true|false) to the querystring.
+/*
+debug isLocalhost result
+true  true        true
+true  false       true
+false true        false
+false false       false
+unset true        true
+unset false       false
+*/
+const boolMap = {
+  'true': true,
+  'false': false,
+};
+const bool = (s) => boolMap[s] ?? null;
+const isLocalhost = window.location.hostname === 'localhost';
+const debugQueryString = () => bool(jHash.val('debug'));
+const DEBUG = debugQueryString() ?? isLocalhost;
+
 $(document).ready(function() {
 
   var matchBodyColor = true;
@@ -116,7 +136,7 @@ $(document).ready(function() {
   $("#searchbox").on('search',search)
   $("#search").click(search)
   $("#customizeChar").on('submit',function(e) {
-    search()
+    search();
     e.preventDefault()
   })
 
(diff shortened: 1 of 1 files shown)


jrconway3 (changes requested):
This is supposed to suppress errors on master, right? It suppresses errors on master and only shows them on localhost.

Well, its still happening here:
https://jrconway3.github.io/Universal-LPC-Spritesheet-Character-Generator/#?body=Body_color_light&head=Human_male_light&hat=Misc_Magic_Hats_starry

However, I did see this warning in the console as well:
> Error with Permissions-Policy header: Unrecognized feature: 'interest-cohort'.

Gaurav0:
This does not suppress File Not Found 404 errors on images. I did some research, and it may not be possible:

https://stackoverflow.com/questions/9893886/prevent-image-load-errors-going-to-the-javascript-console

jrconway3:
So the chosen answer there links to this:
https://stackoverflow.com/questions/7035466/check-if-file-exists-but-prevent-404-error-in-console-from-showing-up

The chosen answer here then says this:
> Years after the question, the answer is finally [yes](https://stackoverflow.com/a/75848002/343721). – 
[Jan Turoň](https://stackoverflow.com/users/343721/jan-turo%c5%88)
 [Commented Mar 26, 2023 at 15:36](https://stackoverflow.com/questions/7035466/check-if-file-exists-but-prevent-404-error-in-console-from-showing-up#comment133789763_7041565)
 
 
The link you listed was from 2017. That being said, its probably not worth the effort just yet. Seems like its pretty complicated to implement.


We could, however, clear the console right away to solve this issue.


Anyway, I still don't know how to properly debug this to ensure the logs you're reporting only happen on master and not local.

Gaurav0:
While this PR does nothing about 404 not found error messages, it shows/hides warnings and logs. The best way to see these is to open the console and compare.

jrconway3:
Right, but I don't know what to look for exactly. I tested missing assets on both and it was the same. I would need to figure out how to accurately replicate when a console error is outputted to see where the error doesn't happen.

jrconway3 (approved):
Okay, so I wasn't looking at the right thing. These are the logs this is addressing:
> Loaded all 71 of 71 assets
> chargen.js:544 Start drawItemsToDraw


They are showing locally but not on master. So this is working.

sanderfrenken:
> While this PR does nothing about 404 not found error messages, it shows/hides warnings and logs. The best way to see these is to open the console and compare.

that is what I was afraid of of, its hard to hide those specifically. I do think we should add a list then to the sheet definitions in case there are animations that contain no image. Something like: 

```

"ignoreImageForAnimations" : ["climb", "sit", ....]

```

And then in chargen we can check this and actually just not load these images. 

For the scope of this however, this PR is fine ofc

PR #88, opened 2025-01-19
* Issue #17 Added Plate Legs to Idle, Climb, Jump, Sit, and Emotes

jrconway3:
For https://github.com/jrconway3/Universal-LPC-Spritesheet-Character-Generator/issues/17

I finished these at the end of last month but didn't want to add them into the sheet until I finished run and revised combat as well. However, now that we've got the assets split I'm able to do a partial update!

Please let me know how these look.


bluecarrot16:
Looks fine to me. I appreciate how much of a massive effort it is to add all these animation frames for legs (which have to be redrawn and can't just be copy-pasted like the heads/hairs/hats). I'm sure after these commits you do as well :p 

(This is why I resisted adding more animations, because it's such a monumental effort to drawn them for every asset...)

jrconway3:
I know its a big effort. I'm doing it mostly because I want to use additional animations myself, mostly run, jump, and climb. The new combat animations were mostly just a bonus... but I did think it could be neat if there were more combat animation options.

Gaurav0 (approved):

Issue #85, opened 2025-01-18
Missing Previews


jrconway3:
A handful of previews are missing from the left side.

I'm mostly seeing this in Backpacks (Torso > Backpack) and Tools/Weapons. For the most part, I think this is happening where an oversize animation is included and no base animations are. However, this is not true for Backpacks, which I don't think have oversize animations at all.

1. Backpack (Torso > Backpack > Backpack)
2. Square Pack (Torso > Backpack > Square Pack)
3. Jetpack (Torso > Backpack > Jetpack)
4. Jetpack Fins (Torso > Backpack > Jetpack Fins)
5. Rod (Tools > Rod)
6. Whip (Tools > Whip)
7. Arrow (Weapon > Ranged > Bow > Ammo)
8. Great Bow (Weapon > Ranged > Bow > Great)
9. Normal Bow (Weapon > Ranged > Bow > Normal)
10. Recurve Bow (Weapon > Ranged > Bow > Recurve)
11. Katana (Weapon > Sword > Katana)
12. Scimitar (Weapon > Sword > Scimitar)
13. Longsword Alt (Weapon > Sword > Longsword Alt)
14. Club (Weapon > Blunt > Club)

All Polearms. I think these are happening in PR #71 but I'd rather merge that and fix the previews later:
1. Cane (Polearm > Cane)
2. Spear (Polearm > Spear)
3. Scythe (Polearm > Scythe)
4. Halberd (Polearm > Halberd)
5. Long Spear (Polearm > Long Spear)
6. Dragon Spear (Polearm > Dragon Spear)
7. Trident (Polearm > Trident)

These are weird ones, only CERTAIN colors are broken in the preview, but the assets work just fine:
1. Heater Shield (Weapons > Shield > Heater Shield > Base > coffee / tan / umber only)
2. Heater Shield Trim (Weapons > Shield > Heater Shield > Trim > gold / iron / steel only)
3. Heater Shield Paint (Weapons > Shield > Heater Shield > Paint > the majority are missing)
4. Heater Shield Pattern (Weapons > Shield > Heater Shield > Pattern > roughly half are missing)

This strange issue is NOT happening on the "Revised" version of the shield at all (which has adjusted positions on some base frames; the "Original" Heater Shield keeps the positioning left behind by Bluecarrot16 on the frames above climb.)

Gaurav0:
Ok, for polearms (Which were broken in PR #71):

[...]

PR #89, opened 2025-01-20
Gaurav/fix/polearms left side previews

Gaurav0:
Partial fix for #85 , the polearms that were regression from #71

Screenshot 2025-01-20 at 8 20 24 AM

Deployed this branch to https://gaurav0.github.io/Universal-LPC-Spritesheet-Character-Generator


sources/chargen.js
```diff
@@ -935,9 +936,13 @@ $(document).ready(function() {
     setTimeout(nextFrame, 1000 / 8);
   }
 
-  function updatePreviewLink(imageLink) {
+  function updatePreviewLink(imageLink, customWalkAnimation) {
     const { directory, file } = splitFilePath(imageLink);
-    imageLink = `${directory}/walk/${file}`;
+    if (customWalkAnimation) {
+      imageLink = `${directory}/${file}`
```
Gaurav0, line 942: This special case is used for animations where there is a walk but we need to preview from walk_128 because walk doesn't have the oversize weapons.

jrconway3 (approved):
Ran tests on this locally, looks good.

sources/chargen.js
```diff
@@ -872,20 +872,21 @@ $(document).ready(function() {
         const previewToDraw = {};
         const animation =  $(this).data(`layer_1_custom_animation`);
 
-        if($(this).data(`layer_1_${getBodyTypeName()}`) === undefined) {
-          let imageLink = $(this).data(`layer_1_${getBodyTypeName()}`);
+        const bodyTypeName = getBodyTypeName();
+        if ($(this).data(`layer_1_${bodyTypeName}`) === undefined) {
+          let imageLink = $(this).data(`layer_1_${bodyTypeName}`);
```
sanderfrenken, line 877: This logic here. Am I overseeing sth or is this wrong?

It states:

```
if ($(this).data(`layer_1_${bodyTypeName}`) === undefined) {
          let imageLink = $(this).data(`layer_1_${bodyTypeName}`);
```
So that will mean that imageLink will always be undefined.. Do you get that?

Gaurav0: Nope you're right. I missed that.

Gaurav0: See #90 first commit.

Issue #91, opened 2025-01-20
Animation Guides


capriryan:
This project is an invaluable resource for hobby game developers like me. Thank you all for your time and effort!

One area I struggled with is determining which frames relate to which animation, the intended frame sequence, and the timing for animations. For example, the generated sprite sheet has nine frames for the walk animation. I assumed all the frames were a part of the walk cycle and could not determine why my animation looked off. I tried to look at the contributors' original work and found ElizaWy's animation guides on GitHub. After seeing their guide begin with the equivalent of the sprite sheet's second frame, I removed the first frame, which looked much better! Below is one of the guides I'm referencing.

![Image](https://github.com/user-attachments/assets/1760d501-14d6-4b71-93d5-53954cf685c4)
[Link to Guide](https://github.com/ElizaWy/LPC/blob/f07f7f5892e67c932c68f70bb04472f2c64e46bc/Characters/_%20Guides%20%26%20Palettes/Animation%20Guides/Walk%20-%2030%20FPS.gif)

The sprite sheet generator's idle animation only has two frames, and ElizaWy's guide has three, so I started to reverse-engineer the preview component in the sprite sheet generator. I found enough information between frame delays and custom sequences to replicate the animations in the preview, but having a guide would have been much easier. Here are some sequences that were unintuitive to me.

Walk: 1-2-3-4-5-6-7-8
Idle: 0-0-1
Jump: 0-1-2-3-4-1
Emote: 0-0-0-0-0-1-1-1-1-1-2-2-2-2-2
Watering: 0-1-4-4-4-4-5
Combat Idle: 0-0-1
Row 46- 1-Handed Slash: 0-1-2-3-4-5-6
Row 46- 1-Handed Backslash: 0-1-2-3-4-5-7-8-9-10-11-12
Rod Thrust: 0-1-2-3-4-5-4-4-4-5-4-2-3

This information may be obvious to those familiar with game development, and perhaps I've missed this information elsewhere, but I did try to look in as many logical places as possible before looking through the back end. It would be immensely helpful to non-experts like me to provide an animation guide similar to ElizaWy's past guides to get us headed in the right direction.

Again, thank you all for your contributions!

jrconway3:
The original idle animation has only two frames. Eliza added a third frame and either option is a valid option. Its up to your preference which one you prefer. When porting order Eliza's Combat Idle I did the same; I only kept 2 frames.

[...]

PR #95, opened 2025-01-21
DOC: simplest way to expose recommended frame cycle

Gaurav0:
For #91


jrconway3 (approved):
Looks pretty good to me.

Issue #94, opened 2025-01-21
Improve UI for smaller screen sizes (Responsive)


Gaurav0:
Right now the site is difficult to work with in smaller screen sizes, especially mobile.

Ideas for a responsive layout:

1. Right column preview canvas for tablet size screens ( ~ 800 - 1280 px ) would instead be placed between sticky form and customize character form. (Sticky form is one that starts with Download and ends with Licenses).
2. Screens without enough height ( < -800 px height ) would have static rather than sticky positioning for header, animation preview, and sticky form.
3. Everything then goes to a single column at mobile sizes.

As a practical matter, note that the preview canvas will still have to be horizontally scrolled on mobile.

PR #97, opened 2025-01-21
Pinch to zoom in/out

Gaurav0:
Partial implementation of #94

This allows you to

- double click to zoom out previewAnimations and spritesheet (click to zoom in still works).
- pinch to zoom in/out previewAnimations and spritesheet independently of viewport. (Pinching outside of those canvases zooms the whole page).
- tap or double tap emulates click / double click and can be used to quickly reset the zoom relative to viewport.

Note that the TinyGesture library uses passive listeners if supported by the browser as recommended by Google and other browser vendors. As a result, it is not going to be as smooth and responsive as a native app, however, it won't freeze and users are unlikely to see poor performance.

### DIFF
--- a/sources/chargen.css
+++ b/sources/chargen.css
@@ -14,8 +14,16 @@ canvas {
   image-rendering: pixelated;
 }
 
-.zoomed {
+.zoomed-in {
   zoom: 2;
+}
+
+.zoomed-out {
+  zoom: 0.5;
+}
+
+.zoomed-in.zoomed-out {
+  zoom: 1;
 }
 
 body {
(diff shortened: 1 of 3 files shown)


jrconway3 (approved):
So basically you're saying if you double tab on either column, in zooms into that column? If I double tap on the spritesheet, it zooms in there... or if I double tap the left navigation, it zooms in there.

And pinch/unpinch for zoom in/out also works? I thought these worked normally in browsers anyway.


Either way, these are working here from what I can tell.

sanderfrenken (approved):

Gaurav0:
> So basically you're saying if you double tab on either column, in zooms into that column? If I double tap on the spritesheet, it zooms in there... or if I double tap the left navigation, it zooms in there.
> 
> And pinch/unpinch for zoom in/out also works? I thought these worked normally in browsers anyway.
> 
> Either way, these are working here from what I can tell.

Pinch zoom does work _for the whole page_ by default. I just made it work independently on the canvases.

PR #101, opened 2025-01-22
fix bows and arrows previews

Gaurav0:
Partial fix for #85


jrconway3 (approved):
The preview from Netlify doesn't work anymore, but I tested this locally and it looks good.

PR #104, opened 2025-01-22
improve search ui

Gaurav0:
1. Show # of matches when searching.
2. When typing in search, debounced search (500ms). This means to go ahead and search any time you stop typing.
3. Draw previews asynchronously. This means that ui will not block (freeze) while loading previews.

I suggest trying to search for "light" or just "lig" to test.

### DIFF
--- a/sources/chargen.js
+++ b/sources/chargen.js
@@ -61,6 +61,7 @@ $(document).ready(function () {
   };
 
   // Preview Animation
+  let past = Date.now();
   var anim = $("#previewAnimations").get(0);
   var animCtx = anim.getContext("2d");
   var animationItems = [1, 2, 3, 4, 5, 6, 7, 8]; // default for walk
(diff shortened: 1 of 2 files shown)


jrconway3 (changes requested):
I got into the preview initially but it is crashing now.

I do see a number of search results. That's pretty cool.


When searching, however, the preview animation freezes. You say this:
> Draw previews asynchronously. This means that ui will not block (freeze) while loading previews.

But it looks like it _is_ freezing to me during search. Is this what you're talking about?

jrconway3:
Same issue happening locally. It does seem like its pretty quick to start the search, but the preview is still freezing and isn't running asyncronously.

Gaurav0:
@jrconway3 updated

jrconway3 (dismissed):
Seems to be working well now.

PR #106, opened 2025-01-22
Add magic wand from https://opengameart.org/content/lpc-wands

### LINKED ISSUE #100: We used to have wands?
https://opengameart.org/content/lpc-wands

Not sure whatever happened to these.

sanderfrenken:
closes #100


jrconway3 (changes requested):
Preview on left side is broken... can we fix that first?

sanderfrenken:
> Preview on left side is broken... can we fix that first?

good catch, fixed!

jrconway3 (approved):
Working now, looks good!

Issue #107, opened 2025-01-23
Getting 529 errors today while searching


Gaurav0:
Getting 529 (Too many requests) errors while testing search today in production (on github.io)

jrconway3:
Using gold, light, dark, I cannot reproduce this issue even on master. So there's no possible way I can review this PR. If I can't even reproduce the issue to begin with, then there's nothing I can do to confirm the PR does anything.

PR #108, opened 2025-01-23
fix 429 errors while searching

Gaurav0:
For #107

### DIFF
--- a/sources/chargen.js
+++ b/sources/chargen.js
@@ -140,7 +140,7 @@ $(document).ready(function () {
       parents.prev("span").addClass("expanded").removeClass("condensed");
       for (const parent of parents.toArray().reverse()) {
         $(parent).delay(50).show().map((i, el) => {
-          setTimeout(() => drawPreviews.call(el), 50 * i);
+          setTimeout(() => drawPreviews.call(el, true), 200 * i);
         });
       }
     }
@@ -951,7 +951,7 @@ $(document).ready(function () {
     imageLoadDone();
   }
 
-  async function getImage2(imgRef, callback, layers, prevctx) {
+  async function getImage2(imgRef, callback, layers, prevctx, fromSearch) {
     if (imgRef && images[imgRef] && images[imgRef].complete) {
       callback(layers, prevctx);
       return images[imgRef];
@@ -967,9 +967,9 @@ $(document).ready(function () {
       await new Promise((resolve) => {
         img.addEventListener('load', function () {
           callback(layers, prevctx);
-          resolve();
+          setTimeout(resolve, fromSearch ? 50 : 4);
         });
-        img.addEventListener('error', resolve);
+        img.addEventListener('error', () => setTimeout(resolve, 50));
       });
       return img;
     }
@@ -984,7 +984,7 @@ $(document).ready(function () {
     }
   }
 
-  async function drawPreviews() {
+  async function drawPreviews(fromSearch = false) {
     const buttons = $(this).find("input[type=radio]")
       .filter(function () {
         return $(this).is(":visible");
@@ -1081,7 +1081,7 @@ $(document).ready(function () {
         });
 
         for (const layer of layers) {
-          img = await getImage2(layer.link, callback, layers, prevctx);
+          img = await getImage2(layer.link, callback, layers, prevctx, fromSearch);
         }
 
         if (img && !$(button).parent().hasClass("hasPreview")) {
@@ -1093,7 +1093,7 @@ $(document).ready(function () {
             .addClass("hasPreview");
         }
 
-        await setTimeout(() => Promise.resolve(), 4);
+        await setTimeout(() => Promise.resolve(), fromSearch ? 50 : 4);
       }
     }
   }


Gaurav0:
Deployed to https://gaurav0.github.io/Universal-LPC-Spritesheet-Character-Generator for testing in a similar environment to production.

Gaurav0:
Still getting 429s. Any ideas?

sanderfrenken:
I didnt try the new search yet. But now I did. I typed gold and hit enter and this happened:

Scherm­afbeelding 2025-01-23 om 19 50 40

I think what happens is that when you search something like "gold", it finds 433 matches. it will expand all these categories at once, and then will try to load all previews associated. So that is like 433*10 = 4300 requests all being fired in rapid succession. That is like a DDOS.

sanderfrenken:
Next to that, it also crashes my browser when I search for EG gold or dark (many results). I am not sure if this was already present in the previous implementation. I never used this functionality really. But I think the search should be restricted to for instance an item name, and dont match on variants in order to prevent opening up 433 lists with each x-amount of previews to draw.

Gaurav0:
> Next to that, it also crashes my browser when I search for EG gold or dark (many results). I am not sure if this was already present in the previous implementation. I never used this functionality really. But I think the search should be restricted to for instance an item name, and dont match on variants in order to prevent opening up 433 lists with each x-amount of previews to draw.

Please test the latest https://gaurav0.github.io/Universal-LPC-Spritesheet-Character-Generator, which I've updated about 3 hours ago. I've only slowed things down, but it seems to be enough.

Gaurav0:
> Next to that, it also crashes my browser when I search for EG gold or dark (many results). I am not sure if this was already present in the previous implementation. I never used this functionality really. But I think the search should be restricted to for instance an item name, and dont match on variants in order to prevent opening up 433 lists with each x-amount of previews to draw.

The previous implementation used to freeze the browser, and as a result I don't think enough requests would fire to trigger the rate limit.

sanderfrenken:
> > Next to that, it also crashes my browser when I search for EG gold or dark (many results). I am not sure if this was already present in the previous implementation. I never used this functionality really. But I think the search should be restricted to for instance an item name, and dont match on variants in order to prevent opening up 433 lists with each x-amount of previews to draw.
> 
> The previous implementation used to freeze the browser, and as a result I don't think enough requests would fire to trigger the rate limit.

yeah that is a very likely explanation

sanderfrenken:
> > Next to that, it also crashes my browser when I search for EG gold or dark (many results). I am not sure if this was already present in the previous implementation. I never used this functionality really. But I think the search should be restricted to for instance an item name, and dont match on variants in order to prevent opening up 433 lists with each x-amount of previews to draw.
> 
> Please test the latest https://gaurav0.github.io/Universal-LPC-Spritesheet-Character-Generator, which I've updated about 3 hours ago. I've only slowed things down, but it seems to be enough.

I searched for "gold" and now it froze my browser

jrconway3:
> But I think the search should be restricted to for instance an item name, and dont match on variants in order to prevent opening up 433 lists with each x-amount of previews to draw.

I'm in agreement with this. I think it should only search by asset name, not variant name.


That being said...


I still can't reproduce this issue. I thought I'd try and do it exactly the way Sanderfrenken did--type gold and press "enter". Still works. It loads a ton of stuff, but I don't get any errors (other than some ACTUALLY missing images).

I thought maybe I wasn't doing it right by just relying on typing and letting it autoload, so I hit "enter" like Sanderfrenken explained, but I'm still not getting the issue. So perhaps its best for you two to test this instead.

sanderfrenken:
> I think matching on variants is useful functionality I don't want to remove.

I think you wont get the behavior right when matching on variant. Usually one would build an API with lazy loading features etc around to prevent the behavior we are seeing at the moment. But that would be quite complex here. You could also just limit a query to 10 results or sth. 

I mean nobody is really helped with opening 433 categories for instances when searching for sth like gold. That is just way too many results to be useful anyways.

sanderfrenken:
> I still can't reproduce this issue.

That is funny. Well maybe your browser has some built-in DDOSing prevention, and it ratelimits by itself the requests sent out. Just a thought. I am on Mac using Safari.

Gaurav0:
> I think matching on variants is useful functionality I don't want to remove. However, perhaps we could consider not drawing previews automatically after searching.

@jrconway3 What do you think of searching for variants but not drawing the previews automatically? A user might have to close and reopen a category to see the previews.

@sanderfrenken I don't think freezing / crashing browsers is great either. Maybe the issue is your computer's memory. How much do you have? I can set a linux vm to that amount and test.

sanderfrenken:
> > I think matching on variants is useful functionality I don't want to remove. However, perhaps we could consider not drawing previews automatically after searching.
> 
> @jrconway3 What do you think of searching for variants but not drawing the previews automatically? A user might have to close and reopen a category to see the previews.
> 
> @sanderfrenken I don't think freezing / crashing browsers is great either. Maybe the issue is your computer's memory. How much do you have? I can set a linux vm to that amount and test.

I dont think it will be the memory. I have 32GB of ram. I quickly checked on Chrome, there it doesn't freeze. So it is also dependent on the browser in use. I do see in Chrome a huge load of 429's though on your site (https://gaurav0.github.io/Universal-LPC-Spritesheet-Character-Generator/#?body=Body_color_light&head=Human_male_light) when searching for gold.

This problem will only grow when adding new assets btw.

jrconway3:
> > I think matching on variants is useful functionality I don't want to remove. However, perhaps we could consider not drawing previews automatically after searching.
> 
> @jrconway3 What do you think of searching for variants but not drawing the previews automatically? A user might have to close and reopen a category to see the previews.
> 
> @sanderfrenken I don't think freezing / crashing browsers is great either. Maybe the issue is your computer's memory. How much do you have? I can set a linux vm to that amount and test.

The reason why it draws the previews is because it opens all the subcategories. We don't really need to open all the subcategories.

sanderfrenken:
> The reason why it draws the previews is because it opens all the subcategories. We don't really need to open all the subcategories.

That is indeed worth a try as well.

Gaurav0:
Closed in favor of #112

PR #109, opened 2025-01-23
feature: tablet view 800 - 1280 px

Gaurav0:
Partial implementation of #94

### DIFF
--- a/sources/chargen.css
+++ b/sources/chargen.css
@@ -37,6 +37,8 @@ header {
   position: sticky;
   top: 0;
   left: 0;
+  margin-top: -10px;
+  padding-top: 10px;
   width: 100%;
   display: flex;
   flex-direction: row;
@@ -152,6 +154,10 @@ input[type=button], input[type=reset] {
   grid-template-columns: 40% auto;
 }
 
+#credits, #advanced {
+  margin-left: 15px;
+}
+
 /* tablet view 800 - 1280 px */
 @media screen and (max-width: 1280px) {
   header {
@@ -160,11 +166,11 @@ input[type=button], input[type=reset] {
 
   #customizeChar {
     display: flex;
-    flex-direction: column;  
+    flex-direction: column; 
   }
 
   #preview {
-    margin-top: 20px;
+    margin-top: 20%;
     max-height: none;
   }
 
@@ -177,11 +183,28 @@ input[type=button], input[type=reset] {
   }
 
   #controls {
-    top: 300px;
+    position: fixed;
+    z-index: 4;
+    margin-top: -10px;
+    padding-top: 30px;
+    padding-left: 20px;
+  }
+
+  /* override #customizeChar > section */
+  #customizeChar > #controls {
+    width: 100%;
   }
 
   #preview-animations {
     margin-left: 15px;
+    position: relative;
+    top: 240px;
+    z-index: 2;
+    opacity: 1;
+    background-color: white;
+    margin-top: -50px;
+    padding-top: 40px;
+    width: 100%;
   }
 }
 
@@ -198,7 +221,6 @@ input[type=button], input[type=reset] {
 ul {
   list-style-type: none;
 }
-
 
 #chooser>ul {
   border-left: solid 1px silver;
@@ -223,7 +245,6 @@ ul {
 .condensed:hover, .expanded:hover {
   text-decoration: underline;
 }
-
 
 .search-result {
   background-color: yellow;
--- a/sources/chargen.css
+++ b/sources/chargen.css
@@ -52,6 +52,11 @@ header > section {
   width: 50vw;
 }
 
+#header-left {
+  margin-top: 10px;
+  margin-left: 15px;
+}
+
 h1 {
   margin: 10px 0px;
 }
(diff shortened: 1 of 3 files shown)


jrconway3:
I haven't had a chance to check this yet. I'll look it over later this evening.

jrconway3 (changes requested):
The "Preview" animation section still remains on the right. (I'm testing scaling down a regular browser, not directly on mobile.) Other than that this looks pretty good overall.

I do worry, however, that the scrolling will be iffy if using an actual tablet.

jrconway3 (changes requested):
So I finally realized I should just use Toggle Device views in Brave.

This is using iPad Air (1180x820 landscape):
image

This looks "kinda" okay, though I feel like the Preview should be closer to the actual sheet itself. However, there are SERIOUS problems with this implementation.

Here:
image


You can BARELY see it. This is... unusable on a mobile device. The buttons up top CANNOT remain fixed like this in this view. On top of that, these buttons are some of the least important things anyway. Its hogging up half the screen which would be better used for other things.

The one thing I think is REASONABLE to remain fixed like this is the preview animation itself. If anything remains in a fixed position it should be that. That way, you can toggle the animations and toggle assets on the animations while still being able to see what you're doing.


In fact, I do wonder if maybe we should actually have the full sheet slide in only when it is toggled. Speaking of which, responsive views almost always do this anyway; a responsive view tends to have the navigation invisible by default, and it only appears when you press a button to toggle it out.


The filters are way too low on the page:
image


These should either be a slide-in navigation or they should be above the image. Honestly, like I said, I think the LEAST useful mechanic to keep on the mobile view is the full spritesheet itself.

If anything, I think it makes more sense to have the animations always visible at all times; its lightweight (in terms of size), you can quickly toggle between other animations, and you can still see what is being toggled. However, this is an option; either way, we should not have both the selection of the assets and the spritesheet above and below each other, and if ANYTHING is at the bottom, the spritesheet makes more sense.


Massive amount of empty space below the filters:
image


Credits WAY too wide:
image


The credits should have a max-width (at least on lower browser sizes) so it automatically adjusts to the size of the screen. Right now its forcibly adding an unnecessary scrollbar which only makes things more confusing and awkward.


When toggling to portrait view:
image


I really don't like how the animations are visible behind the buttons. Those buttons are ridiculously wide as is, going well beyond what they need to take up int terms of width. But again, they don't need to be fixed anyway. Its fine on larger screen sizes because they're not wasting that much space, but here they are wasting a TON of space and aren't necessary.

Gaurav0:
@jrconway3 Thank you for the in depth review. Updated!

jrconway3 (changes requested):
This is an improvement, but it only really addressed the most basic issues. The Preview is in an excellent spot and the credits look good. I think this is slightly more usable...

But most of the major issues I mentioned are still here.


Here's the first view, where I discuss the appearance of the spritesheet:
responsible-tablet-view-comments1

Issues:
1. The buttons at the top of the page take up too much space and I don't see why its so important to make them fixed to the page. Why are they so important? They take up HALF the page.
2. Unnecessary whitespace below the "Preview" section.
3. Minor issue, but I think we shouldn't be required to scroll passed the entire spritesheet to get to the filters. In fact, I don't think the spritesheet needs to appear by default. However, addressing issues #1 and #2 I think will be fine if you don't want to address this one.


Southern half:
responsible-tablet-view-comments2

Issues::
1. I have screenshotted the full screen this time. (I'm using F12 in Brave, "Toggle Device" button to get tablet view.) The Preview appears about halfway down the page, pushed down by the "Download" / "Import" / etc buttons. Why are these necessary here? They should either be static at the top of the page, or they should be a pull out.
2. Something I forgot to mention, but the Licenses really should be with the Filters. I didn't catch that they're actually part of this upper section before.
3. Again, the whitespace is showing major problems here, especially the whitespace at the bottom of the filters. This is INSANELY long for no good reason.

Gaurav0:
Once again, thanks for the great feedback. I'm going to continue to work at it to get it right.

Gaurav0:
Fixed the whitespace below the customize section and above the credits. It was some obsolete javascript code causing that.

Gaurav0:
> 3\. Minor issue, but I think we shouldn't be required to scroll passed the entire spritesheet to get to the filters. In fact, I don't think the spritesheet needs to appear by default. However, addressing issues [Updated Female Hairstyles and Colors #1](https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/pull/1) and [Tiaras, Bandanas, Sub Categories Added #2](https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/pull/2) I think will be fine if you don't want to address this one.

Actually I do. But I don't think it should be hidden by default. I think in a followup PR, I should create an accordion UI where most sections should be collapsible / expandable. This is an accordion: https://jqueryui.com/accordion/

jrconway3 (approved):
Okay, I don't think it has to be perfect. This is definitely a much better and more usable editor on tablet.


I still think its a good idea to have the Preview animation shift to being a fixed asset as you scroll down, but that's hardly a requirement or anything. However, if "Preview" were to be converted into a fixed asset when scrolling down, that would make the filters far easier to use.

jrconway3:
> The controls section, which currently includes the licenses, is not fixed unless the height is 1000px or higher, which is usually "portrait mode". I'm going to increase this threshold, but honestly this section really is important. As you know, we get way too many questions about licenses, how to download, etc. Did you notice there is no other download button on the page? Anyway, I think it needs to be prominent.

I don't see why it needs to be fixed. We can leave it at the top of the page, so that way it is prominent. But why exactly does it need to be fixed so its visible at all times?


I didn't complain about this on the other view because it doesn't take up that much space (though I still don't think it needs to be fixed). However, on mobile especially it should not be fixed because it takes up way too much screen real estate.


If you want ANYTHING to be fixed, perhaps just a download button would be enough. Or again, we can have a pullout menu. It could even convert into a pullout menu as you scroll down.



We're already doing a pseudo accordion already with the filters, though I do approve of other accordion features on mobile. That being said, my thought is still having the filters and the spritesheet side-by-side rather than being an accordion. So we have buttons that slide the spritesheet back and forth.


Thanks for confirmation on the Preview, I get why it needs to be larger now.



Anyway, this is much better now that the controls aren't fixed and the whitespace at the bottom is gone. The controls being fixed to the page took up way too much room.

One thing that can solve this problem, like I mentioned, is the Licenses filter being put into the Filters section instead. Being available up top by default is taking up a lot of extra room that isn't necessary. On mobile devices we have far less real estate to work with than on desktop.

Gaurav0:
> I still think its a good idea to have the Preview animation shift to being a fixed asset as you scroll down, but that's hardly a requirement or anything. However, if "Preview" were to be converted into a fixed asset when scrolling down, that would make the filters far easier to use.

Actually I tried that to see how it would look. For some reason, applying `position: sticky` on `#preview-animations` in the tablet view media query isn't working in Chrome.

Gaurav0:
> One thing that can solve this problem, like I mentioned, is the Licenses filter being put into the Filters section instead. Being available up top by default is taking up a lot of extra room that isn't necessary. On mobile devices we have far less real estate to work with than on desktop.

I've opened a separate PR for this.

PR #120, opened 2025-01-26
Gaurav/feat/move licenses into customize

### LINKED ISSUE #119: Move Licenses Selection Into Filters
Licenses should not be part of the fixed upper section, they should be in the Filters and HIDDEN by default (like Supported Animation).

Here's the current appearance:
Image

Here's what it should look like:
Image

This is haphazard, I can't get the selector to work when manually moving things around in Developer Tools. But basically, I don't see any reason why we need this to be in the upper section or visible by default. Just show the default licenses, and if users WANT to toggle licenses, they can. However, its closed by default to put more emphasis on the assets, just like how Supported Animations works.

I also think that "Supported Animation" probably shouldn't be in "Body". Either we can add a "Meta" header or something else, maybe something better descriptive. Perhaps just say "Filters" or something. I put that into my example above.

Gaurav0:
Based on #109 - please review that PR first.

Fixes #119

### DIFF
--- a/sources/chargen.css
+++ b/sources/chargen.css
@@ -131,6 +131,7 @@ input[type=button], input[type=reset] {
   z-index: 2;
   margin-left: 15px;
   order: -1;
+  height: 60px;
 }
 
 #chooser {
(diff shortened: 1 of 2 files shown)


jrconway3 (dismissed):
Looks good to me. This also makes the controls section a lot better on desktop as we have more room for the filters.

Issue #122, opened 2025-01-26
Accordion features for tablet/mobile


Gaurav0:
Use an accordion UI layout to make navigating easier on tablet/mobile.

PR #124, opened 2025-01-26
Make sections collapsible in tablet view

Gaurav0:
For #122

### DIFF
--- a/sources/chargen.css
+++ b/sources/chargen.css
@@ -168,10 +168,20 @@ input[type=button], input[type=reset] {
   margin-left: 15px;
 }
 
+summary {
+  display: none;
+}
+
 /* tablet view 800 - 1280 px */
 @media screen and (max-width: 1280px) {
   header {
     flex-direction: column;
+  }
+
+  #header-left {
+    width: calc(100% - 140px);
+    min-height: 180px;
+    overflow-y: auto;
   }
 
   #customizeChar {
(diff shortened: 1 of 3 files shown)


jrconway3 (changes requested):
I can't see anything inherently wrong with this, but IDK. Something feels weird about it. But I can't pinpoint what.


That said, I do have one relatively minor complaint:
image


When hiding the spritesheet there's a fairly large gap of whitespace below it. Its not as bad as the prior sets, but it still looks like there's just too much here IMO.

jrconway3 (approved):
I can't see anything else particularly notable at this time.

jrconway3:
There's a horizontal scrollbar on the tablet view. Why is there a scrollbar?

I wonder if its related to Advanced Tools. When on portrait orientation on tablet the Advanced Tools extends slightly wider than the screen.

Gaurav0:
@jrconway3 Scrollbar should be gone.

jrconway3 (changes requested):
Still happening here:
image

However, it is better.

The bottom is improved:
image

That said, I still think "Z-position" should go down to the next line on small views.


And its no longer happening on landscape:
image



Could this be due to the canvas restrictions because of oversized animations?

Gaurav0:
The minimum width of the canvas is 832 px when there are no oversize animations.

jrconway3:
Can we make the canvas scroll on its own (or just the block the canvas is in)? That way we don't have this "entire page" scrolling feature on mobile which is not user-friendly, and when we hide the spritesheet, the width of the page is the same size as your current browser.

Gaurav0:
@jrconway3 Yes. Done.

jrconway3 (changes requested):
I'm supposed to be asleep ugh... :P


This looks great:
image


...but why is there STILL a scrollbar here:
image


Also, z-index looks great, too.

jrconway3:
Found it, so apparently its this:
```
<canvas id="previewAnimations" width="1024" height="64"></canvas>
```

And this extra space isn't even necessary.

Gaurav0:
@jrconway3 Updated again.

jrconway3 (approved):
Looks good now. No longer having that issue.
image


It even looks decent on mobile, too:
image


Though the collapsible sections is gone for some weird reason. I know its not in scope so its not a blocker, but mobile is almost ready as is.

Gaurav0:
Collapsible sections aren't enabled for "portrait mode" on tablets and retina displays. That's what I'll work on next.

PR #125, opened 2025-01-26
fix customize character previews extend beyond viewport width

Gaurav0:
Fix for bug in #109

Before:

Screenshot 2025-01-26 at 6 25 30 PM

After:

Screenshot 2025-01-26 at 6 26 44 PM


jrconway3 (approved):
Works great from what I can see.

Issue #123, opened 2025-01-26
Controls Layered Over Filters When Scrolling to Bottom of Page


jrconway3:
Was examining the credits to see how it works (wound looks weird due to the split spritesheets now) and that's when I noticed this issue:
![Image](https://github.com/user-attachments/assets/d516c70a-806a-4026-aee2-092a8db4c4aa)

For some reason, once you scroll low enough that the controls start to scroll up on the page, the background is no longer flat white and the filters appear behind the buttons. This doesn't look good. Not sure exactly what would cause this to happen.

PR #126, opened 2025-01-27
fix controls z_index

Gaurav0:
For #123 

It doesn't make sense why this should be necessary, these properties should be inherited. Oh well, it works.


jrconway3 (approved):
looks good

Issue #129, opened 2025-01-27
Add Support for Prerequisite Assets...? (or Toggle Asset Directory if Another Asset Enabled)


jrconway3:
I'm not exactly sure how to do this, but there's a functionality I've been interested in for a while now that may highly benefit overlay assets.

This is in relation to multiple issues:
- #102 
- #66 
- #42 

Its especially important for #102 because I'm actively working on this right now.

Here's the problem I'm seeing:

Eyes overlays only work for specific heads. Of course its all the standard human heads, so that covers a LOT of things. However, what happens when we add Expressions in?

Obviously, the eyes just won't work anymore. We need a new overlay for the eyes in the new positions.

This is pretty easy to do. Once I map the expressions to all heads, I can create a duplicate asset and remove the skintone so we just have the eyes left. Then we can have an eye color swap overlay.

However, we would need to manually toggle the swap PER asset... or perhaps we can add the ability to change the directory paths if specific assets are enabled? Another option: have a default required asset, and if that asset is turned off, the asset goes away.

This second option will work well for many overlay assets, such as what I described in #42:
When we toggle between different hats, the trims just don't work anymore. What I would want to do in such cases is, if the root asset is changed, we remove the trim recolor.

I'm bringing this up here because I'm not exactly sure how to do this, and it will likely become important pretty soon. Here's an example of a toggle in the code for something similar:
```
  "match_body_color": true,
```

Match body color ensures that whatever asset is added always matches the same variant as the body, so it forcibly changes the body color. This is an example of a prerequisite, but it only does one thing--change the body color.

Gaurav0:
Right now, the body matching algorithm unfortunately uses data-required="male,female" attribute.

I wish I had made it data-required="body=male,female" or something.

Gaurav0:
> However, we would need to manually toggle the swap PER asset... or perhaps we can add the ability to change the directory paths if specific assets are enabled? Another option: have a default required asset, and if that asset is turned off, the asset goes away.

- I really want the directory paths to be set in the sheet definitions as much as possible. We can use a JSON object map.
- The second I definitely want to do, regardless of whether we do the first.

I'm thinking now I should:

1. rename `data-required` to `data-sex-required`
2. add special code for a `data-head-required` This is because head type isn't its own key right now.
3. add more general code for `data-required` that allows specifying any `key=<csv list>`
4. generalize the general code to allow multiple keys

[...]

PR #130, opened 2025-01-28
* Issues #102 Added ElizaWy's Expressions

### LINKED ISSUE #102: Eliza's expressions
Incorporate expressions by Eliza.

jrconway3:
There's way, way too many changes here, so its probably best to test through Nelify.

I managed to add Eliza's expressions to all frames except climb (north only) and hurt (it doesn't make sense for most expressions to be on hurt; I did add closed eyes though). This should resolve #102 

Male, Elderly, and Female Expressions: 
1. Closing Eyes
2. Closed Eyes
3. Looking Right
4. Looking Left
5. Eyeroll
6. Anger
7. Sad
8. Happy
9. Blush
10. Shame
11. Shock

Also added eye overlays for all the following frames:
1. Closing eyes
2. Looking Right
3. Looking Left
4. Eyeroll
5. Anger
6. Sad
7. Shame
8. Shock

Lastly, I updated "Default" Eyes:
1. Eyes are now labeled as Default Eyes
2. Child Eyes are a separate category
3. Elderly Eyes also have "Default" Eyes

I have added all these eye sheets as separate overlays, but IMO ideally we'd be able to toggle between each sheet based on the expression, or have the expression itself replace the eye sheet based on color (see #129).

I forgot to replace the eye colors on the expressions before separating all the sheets out... I did manage to fix the default eye overlay colors, though, so they're only using ULPC colors and not LPCR colors. I still added an eye color conversion sheet from LPCR to ULPC though.


Gaurav0:
Could you split this out into 3 commits?

1. html and js changes (if any)?
2. sheet definitions changes
3. new png files

Thanks.

Gaurav0:
I'd like to see in the future, we'd be able to just select an expression and automatically get:

1. expression matching head type
2. expression matching head color
3. expression matching eye color

Gaurav0:
> I forgot to replace the eye colors on the expressions before separating all the sheets out... I did manage to fix the default eye overlay colors, though, so they're only using ULPC colors and not LPCR colors. I still added an eye color conversion sheet from LPCR to ULPC though.

Do we want to still redo this in a separate PR?

jrconway3:
> I'd like to see in the future, we'd be able to select an expression and (perhaps optionally) automatically get:
> 
> 1. expression matching head type
> 2. expression matching head color
> 3. expression matching eye color

I think it depends on the head type. Head type is not really THAT important, though you could argue that we'd want to restrict the expressions only to human heads. However, for the most part all the expressions look good on any human head interchangeably. I think the main difference between the human head types is actually the face. Though the male heads are also slightly differently shaped on the bottom.

Eye Color is what I'm way more concerned with. As I said before, this is something we would need to do by adding a "prerequisite" functionality. That's why I opened the issue #129. I don't really know how to implement that kind of functionality, maybe if I looked real hard and dug into the code I might be able to do it, but I'm more focusing on assets atm.

The head colors already match though. That's already in place; body color matching is the one feature that already exists.





> Could you split this out into 3 commits?
> 
> 1. html and js changes (if any)?
> 2. sheet definitions changes
> 3. new png files
> 
> Thanks.

The first isn't really relevant. The sheet definition changes make the html changes, but that's about it. Most of the time when adding assets the changes are the files themselves and the sheet definitions + source_index.html and the index.html file. The index.html file is generated from source_index.html and sheet definitions. On top of that, PR's break if the html doesn't match, so we really have to drop both at the same time.

I can, however, drop the PNG's separately.




> > I forgot to replace the eye colors on the expressions before separating all the sheets out... I did manage to fix the default eye overlay colors, though, so they're only using ULPC colors and not LPCR colors. I still added an eye color conversion sheet from LPCR to ULPC though.
> 
> Do we want to still redo this in a separate PR?

I suppose we might want to do this. I really should fix them first, but it was such a PITA to get that far as is. What I might be able to do is generate all the eye color variants and pull the correct eye color variant back to the parent directory. I used a little mmv command to do this already for the eye overlays as is.


I'll leave this PR open as a WIP so you and Sanderfrenken can at least test it.

PR #134, opened 2025-01-28
make preview and customize char sections collapsible on portrait mode

Gaurav0:
... for tablets and retina phones

### DIFF
--- a/sources/chargen.css
+++ b/sources/chargen.css
@@ -28,7 +28,7 @@ canvas {
 
 body {
   margin: 10px 15px;
-  font-size: 10pt;
+  font-size: 16px;
   font-family: monospace;
 }
 
@@ -258,6 +258,28 @@ summary {
     margin-top: 0;
     padding-top: 0;
   }
+
+  #preview summary, #chooser summary {
+    display: list-item;
+  }
+
+  #preview details #chooser details {
+    position: relative;
+    top: -4rex;
+  }
+
+  #preview details[open], #chooser details[open] {
+    margin-top: 12rex;
+  }
+
+  #preview:has(+#controls>details:not([open])) {
+    margin-top: calc(170px + 12rex);
+  }
+
+  #preview:has(>details[open]) {
+    margin-top: calc(60px + 12rex);
+    margin-bottom: 60px;
+  }
 }
 
 /* tablet view 800 - 1280 px, <= 1200 px height */
@@ -292,15 +314,6 @@ summary {
     min-width: unset;
   }
 
-  summary {
-    display: list-item;
-  }
-
-  details {
-    position: relative;
-    top: -4rex;
-  }
-
   #controls:has(>details:not([open])) {
     position: relative;
     top: 40px;
@@ -311,6 +324,7 @@ summary {
   }
 
   #preview:has(>details[open]) {
+    margin-top: calc(60px + 12rex);
     margin-bottom: 60px;
   }
 }
--- a/sources/chargen.css
+++ b/sources/chargen.css
@@ -28,7 +28,7 @@ canvas {
 
 body {
   margin: 10px 15px;
-  font-size: 16px;
+  font-size: 10pt;
   font-family: monospace;
 }
 
@@ -72,7 +72,6 @@ h2 {
 }
 
 h3 {
-  font-size: 1em;
   margin: 4px 0px 0px 0px;
 }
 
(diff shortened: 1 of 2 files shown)


sources/chargen.css
```diff
@@ -258,6 +258,28 @@ summary {
...
+    position: relative;
+    top: -4rex;
+  }
+
+  #preview details[open], #chooser details[open] {
+    margin-top: 12rex;
+  }
+
+  #preview:has(+#controls>details:not([open])) {
+    margin-top: calc(170px + 12rex);
```
Gaurav0, line 276: In theory, 1 rem is the width of the letter M in the font, and 1 rex is the width of the letter x in the font. The "r" for "root" indicates that we are talking about the font on the body, not the font on the element.

In practice, on Chrome and Safari, 16 px = 1 rem = 2 rex.

On retina displays, rem and rex units are respected while px units are multiplied by the dip (device/density independent pixel) size, usually 2.

So I'm mixing these units because on those retina displays, 12 rex is going to be 192 px while on regular displays they will be 96 px. This offsets that the body font size (set to 16px now) will be 16px on regular displays but 32px on retina displays.

Anyway, hope that makes sense. I suggest you review by looking at retina phones and tablets using the device picker in Chrome devtools.

jrconway3 (approved):
Apologies for the delay, but let's take a look at this...


The biggest issue I'm seeing is inconsistent text:
image

The text size above of the "Preview" animation is barely readable, but after this, its GIGANTIC. Why is the text sizes so inconsistent?


One confusion here for me is that this view doesn't say anything about it being retina, even though it is an iOS view. However, when I first toggle to this view, it tells me:
980x2120. Not exactly sure what that means, and it doesn't sound like retina because iPhone 12 Pro says its 390 x 844...  IIRC, I thought retina was x2.


iPhone 12 Pro's landscape shows Preview being cut off:
image

Both views of iPad Air are cutting of Preview:
image

Pretty sure iPad Air was working before.

iPad Pro looks fine, though

jrconway3 (changes requested):
I accidentally marked as approved rather than Request Changes. Please see prior comment.

jrconway3:
Here's what it looks like on my Moto G Power:
Screenshot_20250201-133507

And landscape:
Screenshot_20250201-133546


So the portrait view has the same issue as using chromium where the font sizes are inconsistent. However, landscape looks fine.

jrconway3 (changes requested):
This certainly looks a lot better, but now maybe its a bit TOO small...?
image


Originally some text was this small, and some text was ridiculously large. The large text is no longer happening, and it looks more consistent, so that's an improvement.

This looks very consistent now:
image


Its still small, but it might look better on an actual mobile device.

Its still weirdly inconsistent in my Moto G Power... perhaps that's just because its cached?

jrconway3:
Here's what it looks like on my Moto G Power:
Screenshot_20250201-135250


I doubt this is something very easy to fix, it might even be cached or something. I would just recommend making the fonts slightly bigger than we can call of a day on this one

jrconway3 (changes requested):
When I load the view by default, I get this:
image

Its not zooming in by default anymore for some weird reason. It zooms properly if I double tab, but not by default. This happens on mobile, too. It wasn't doing this before, so its... really weird.


This is inconsistent, with small text on some areas:
image


However, when I CLICK on it, it restores to the correct size of the other links:
image

Gaurav0:
@jrconway3 I ended up doing a refactor of the css.

- Removed the header and made everything use a display: grid
- Reordered some html
- Reverted the viewport change and the base font size change
- Created four separate sections to apply css
  - \> 1280px width
  - < 1280px width
  - < 1280 px width and <= 1200 px height
  - < 1280 px width and > 1200 px height

The css is much simpler now, and I'm also no longer seeing font sizing bugs in the devtools phone previews I was seeing before.

Hopefully it will be easier to fix any bugs you find now.

jrconway3 (changes requested):
I want to say "maybe this is just some sort of fluke on my end", but its happening on both my phone and in my Brave browser on my desktop:

Screenshot_20250202-094114
Screenshot_20250202-094124

Literally the exact same issue... the first two sections of the filters are significantly smaller.

jrconway3:
Windows version (Brave):
image


Even Edge:
image

Gaurav0:
@jrconway3 For reference:

https://stackoverflow.com/questions/44378664/rem-based-layouts-zooming-on-chrome-is-inconsistent-px-vs-rem

I am thinking of specifying all font-size in pt.

Gaurav0:
@jrconway3 Updated again. Fonts might be too small on some retina displays now, but at least it isn't all buggy.

jrconway3 (approved):
Everything looks fine now. I checked earlier and something still looked off, and I started leaving a comment, but apparently I forgot to post it. However, I'm not seeing anything now.

Issue #133, opened 2025-01-28
Add Laetissima and Pennomi's Character Expressions


jrconway3:
From here:
https://opengameart.org/content/lpc-base-character-expressions

The majority of these assets are already present, but there's a small handful of things NOT here, notably the crying tears. I'll look through and see if I can find anything else useful that Eliza doesn't already have.

Only doing this now because, well, since we've got Eliza's stuff ready to go from #102, there's no real harm in also including the final stuff from here, too.

I'll do this only after Eliza's expressions get merged first. Once they are merged, I'll add the remaining few things from this old update by Laetissima and Pennomi.

jrconway3:
Here's what's in here:
1. Tears
2. Sad Eyes
3. Angry Eyes
4. Happy Eyes

The tears are the most interesting one by far I think. Sad/Angry/Happy, on the other hand, we do have with Eliza, but I'll export them and see how different they are compared to Eliza's.

jrconway3:
Because I took a step back to redo Eliza's Expressions... I'll just drop these ones at the same time, now.

PR #138, opened 2025-01-30
Issue 102 test issue 129

jrconway3:
Resolves issues #102 and #133 while also implementing Gaurav's fix for #129.

Only thing missing atm is custom assets for Elderly specifically. I replaced the original facial expressions to reduce them down to the minimum number of pixels needed to create the expressions from the base eyes. Because of this, though, I realized it would mess up spellcast (which has the eyes closing), so I dropped spellcast for now until I have a solution for it.

UPDATE:
A new variant PR is here:
https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/pull/138

Please see poll here:
https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/discussions/139

This one updates ElizaWy's assets to work similarly to Pennomi and Laetissima's assets. The only problem with it is that it doesn't replace the full faces so that also means spellcast doesn't work, because spellcast as the eyes opening and closing.


jrconway3:
Can't figure out why the templates aren't working...

jrconway3:
Yeah so looks like its not properly configured yet. If you test out this branch you can see that none of the faces work. That's because its keeping the var inside the URL rather than replacing it:
```
https://deploy-preview-138--liberatedpixelcup.netlify.app/spritesheets/head/faces/look_l/$%7Bhead%7D/run/blue.png
```

`${head}` should be replaced with "adult" or "elderly".

jrconway3:
Merged Gaurav's updates:
https://deploy-preview-138--liberatedpixelcup.netlify.app/#?body=Body_color_light&head=Human_male_light&expression=Closing_Eyes_human_male_light&eyes=none_Eyes


Eyes are messed up and don't work at all:
image


I was concerned about the Faces having the same name across multiple categories, however, that is taken into account:
image


In this URL, it clearing adds "human_male" into the expression section, so it at least does cover that:
https://deploy-preview-138--liberatedpixelcup.netlify.app/#?body=Body_color_light&head=Human_male_light&expression=Closing_Eyes_human_male_light&eyes=none_Eyes

Issue #148, opened 2025-02-03
tags / excluded by


Gaurav0:
Split from https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/issues/129#issuecomment-2620240517

@BenCreating  wrote:

> I implemented this in [my generator](https://github.com/BenCreating/LPC-Spritesheet-Generator) with the tags and excluded-by properties in the spritesheet definition. With that system a head asset would be given the tag human or animal, and the expression asset would list animal in the exclusions. In most cases, you only need to edit the definition for the asset you're adding.

> Admittedly, with my system you would need to update the expression definitions if you added a new incompatible (non-animal) head type, but I'm assuming whoever creates/adds that asset will be more aware that additional changes need to be made to prevent incorrect spritesheets.

PR #149, opened 2025-02-03
Implement tags/required_tags/excluded_tags

Gaurav0:
For #148

### DIFF
--- a/sheet_definitions/backpack_contents_ore.json
+++ b/sheet_definitions/backpack_contents_ore.json
@@ -1,6 +1,8 @@
 {
   "name": "Ore",
   "type_name": "cargo",
+  "tags": ["backpack_contents"],
+  "required_tags": ["backpack"],
   "layer_1": {
     "zPos": 140,
     "male": "backpack/basket_contents/ore/fg/",
--- a/sheet_definitions/cape_solid.json
+++ b/sheet_definitions/cape_solid.json
@@ -1,6 +1,8 @@
 {
   "name": "Solid",
   "type_name": "cape",
+  "tags": ["cape"],
+  "excluded_tags": ["backpack", "wings"],
   "layer_1": {
     "zPos": 85,
     "male": "cape/solid/female/",
--- a/sheet_definitions/cape_tattered.json
+++ b/sheet_definitions/cape_tattered.json
@@ -1,6 +1,8 @@
 {
   "name": "Tattered",
   "type_name": "cape",
+  "tags": ["cape"],
+  "excluded_tags": ["backpack", "wings"],
   "layer_1": {
     "zPos": 85,
     "male": "cape/tattered/female/",
--- a/sheet_definitions/shield_heater_revised_trim.json
+++ b/sheet_definitions/shield_heater_revised_trim.json
@@ -1,7 +1,7 @@
 {
   "name": "Revised Heater Shield Trim",
   "type_name": "shield_trim",
-  "excluded": ["heater"],
+  "required_tags": ["heater"],
   "layer_1": {
     "zPos": 6,
     "male": "shield/heater/revised/trim/bg/",
--- a/sheet_definitions/shield_heater_revised_trim.json
+++ b/sheet_definitions/shield_heater_revised_trim.json
@@ -1,6 +1,7 @@
 {
   "name": "Revised Heater Shield Trim",
   "type_name": "shield_trim",
+  "excluded": ["heater"],
   "layer_1": {
     "zPos": 6,
     "male": "shield/heater/revised/trim/bg/",
--- a/sheet_definitions/shield_heater_wood.json
+++ b/sheet_definitions/shield_heater_wood.json
@@ -1,6 +1,7 @@
 {
   "name": "Heater Shield Base",
   "type_name": "shield",
+  "tags": ["heater"],
   "layer_1": {
     "zPos": 2,
     "male": "shield/heater/original/wood/bg/",
--- a/sources/chargen.js
+++ b/sources/chargen.js
@@ -877,6 +877,8 @@ $(document).ready(function () {
     const bodyType = getBodyTypeName();
     const selectedAnims = getSelectedAnimations();
     const allowedLicenses = getAllowedLicenses();
+    const promises = [];
+    const lists = new Set();
 
     // only interested in tags if on a selected item
     const selectedTags = new Set();
@@ -988,7 +990,8 @@ $(document).ready(function () {
 
       // Display Result
       if (display) {
-        $this.show();
+        promises.push($this.show().promise());
+        lists.add($this);
       } else {
         $this.hide();
       }
@@ -1040,7 +1043,8 @@ $(document).ready(function () {
 
       // Display Result
       if (display) {
-        $this.parent().show();
+        promises.push($this.parent().show().promise());
+        lists.add($this);
       } else {
         $this.parent().hide();
       }
@@ -1056,6 +1060,14 @@ $(document).ready(function () {
       $(".removeIncompatibleWithLicenses").show();
     } else {
       $(".removeIncompatibleWithLicenses").hide();
+    }
+
+    if (promises.length > 0) {
+      Promise.allSettled(promises).finally(() => {
+        for (const $li of lists) {
+          drawPreviews.call($li.get(0));
+        }
+      });
     }
   }
 
--- a/sources/chargen.js
+++ b/sources/chargen.js
@@ -579,7 +579,7 @@ $(document).ready(function () {
   function whichPropChecked(ids, key, vals) {
     const regExps = vals.map(val => new RegExp(String.raw`^${key}-${val}`, "i"));
     const els = findIdsByRegExp(ids, regExps);
-    for (const i = 0; i < vals.length; ++i) {
+    for (let i = 0; i < vals.length; ++i) {
       if (els[i] === true) {
         return vals[i];
       }
@@ -886,12 +886,14 @@ $(document).ready(function () {
         selectedTags.add(tag)
       );
     });
-    console.log('selectedTags', selectedTags);
 
     let hasUnsupported = false;
     let hasProhibited = false;
 
-    $("#chooser li").each(function (index) {
+    $("#chooser li[data-required]").each(function (index) {
+      let hasExcluded = false;
+      let excludedText = '';
+
       // Toggle Required Body Type
       const $this = $(this);
       const dataRequired = $this.data("required");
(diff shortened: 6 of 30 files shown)


sources/chargen.js
```diff
@@ -877,6 +877,16 @@ $(document).ready(function () {
...
+
+    // only interested in tags if on a selected item
+    const selectedTags = new Set();
+    $("input[type=radio]:checked").each(function () {
+      const tags = $(this).data("tags");
+      tags && tags.split(",").forEach(tag =>
+        selectedTags.add(tag)
+      );
+    });
+
```
jrconway3, line 889: I don't see anything wrong with this, but we do need to be careful with tags if we're making them global.

jrconway3 (changes requested):
Left some comments. I'll address situations where `excluded` can be used in:
https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/issues/148

sources/chargen.js
```diff
@@ -967,6 +977,14 @@ $(document).ready(function () {
         }
       }
 
+      // Toggle based on tags/excluded
+      const excluded = $this.data("excluded");
+      excluded?.split(",")?.forEach(tag => {
+        if (selectedTags.has(tag)) {
+          display = false;
+        }
+      });
```
jrconway3, line 986: Should be required, rather than excluded. Though we could label them as:
```
required_tags
excluded_tags
```


The structure here can be duplicated for required tags, just need to reverse it so that if required_tags is added, all tags are required for the asset to work.

sheet_definitions/shield_heater_wood.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "Heater Shield Base",
   "type_name": "shield",
+  "tags": ["heater"],
```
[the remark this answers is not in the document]
Gaurav0: While you're probably right, I'm more interested in testing the mechanism right now. I'll leave this to fix in a later commit or followup PR.

sources/chargen.js
Gaurav0: Correct.

sheet_definitions/backpack_contents_ore.json
```diff
@@ -1,6 +1,8 @@
 {
   "name": "Ore",
   "type_name": "cargo",
+  "tags": ["backpack_contents"],
+  "required_tags": ["backpack"],
```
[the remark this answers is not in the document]
jrconway3: Okay, so there IS a problem here:
image

If you toggle on the backet, then the cargo is enabled. If you toggle OFF of the basket, the cargo is not only still enabled and visible, it doesn't disappear from the sheet at all.

sheet_definitions/cape_solid.json
```diff
@@ -1,6 +1,8 @@
 {
   "name": "Solid",
   "type_name": "cape",
+  "tags": ["cape"],
+  "excluded_tags": ["backpack", "wings"],
```
jrconway3, line 5: image


Like I said before, I think this is fine, though my preference would be to disable the category in some more notable way.

Previously I said we should hide the category entirely, but perhaps instead we should say why it doesn't work:

"incompatible with X" for example. Say what asset it doesn't work with to make it clearer what the user needs to disable to allow it to work.


(Personally I would not restrict cape if wings or backpack are enabled, but I understand this is just an example.)

sheet_definitions/shield_heater_revised_trim.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "Revised Heater Shield Trim",
   "type_name": "shield_trim",
+  "required_tags": ["heater"],
```
jrconway3, line 4: image


So... this is probably fine, I think. I was going to leave a comment that maybe we should say "requires asset X", but that would be far more complicated. If we are simply saying that an asset cannot be added because of another asset, that's easier, because we can check in the current selection what assets are selected that's acting as a restriction.

For this, though, we'd have to list off the names of all assets that have the tag, because we can't use the tag to reference it. I think as long as the assets are next to each other, its probably fine.

sheet_definitions/backpack_contents_ore.json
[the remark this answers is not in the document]
jrconway3: Correction:
image

As it turns out, the jetpacks aren't configured to lock out the cargo. Once I disable the jetpack, then the cargo items disappear. So that IS working correctly.


However, its only removing the items from the list. The asset still appears on the right side. When disabling the prerequisite asset, we should also remove the asset from the sheet.

sheet_definitions/cape_tattered.json
```diff
@@ -1,6 +1,8 @@
 {
   "name": "Tattered",
   "type_name": "cape",
+  "tags": ["cape"],
+  "excluded_tags": ["backpack", "wings"],
```
jrconway3, line 5: image


Okay, so, there's one weird issue here. The capes work fine once you disable backpack or wings. However, if you had the category open and the assets get filled out, it doesn't generate the previews.

And of course, you can see that the "cargo" is still appearing on the right side, too, even after I removed its prerequisite asset.

jrconway3 (changes requested):
Left a few comments.

One thing I have noted in these comments is that when an asset is disabled, there's no indication as to what asset needs to be enabled OR disabled to make it work.


I'm thinking, perhaps it should be something like this:

#### Included Assets:
For an "include" asset, the entire category should only appear if the pre-requisite asset is enabled. This can be seen in "backpack cargo" and "heater shield trim."

#### Excluded Assets:
For an "exclude" asset, once an asset that excludes it is enabled, we should list off what asset is blocking it.

For example, if you enable a specific cape, we can grab the cape from our asset selection and show in the category "This cannot be used with X" where X = the current selected asset.

1. Choose a Dragonfly Wing Asset
2. Open the Capes Category
3. Where it says "no cape", instead show "Cape cannot be used with Dragonfly Wing"


These two things are optional, but I think they would make this feature easier to use.

Gaurav0:
> Left a few comments.
> 
> One thing I have noted in these comments is that when an asset is disabled, there's no indication as to what asset needs to be enabled OR disabled to make it work.
> 
> I'm thinking, perhaps it should be something like this:
> 
> #### 1. Included Assets:
> For an "include" asset, the entire category should only appear if the pre-requisite asset is enabled. This can be seen in "backpack cargo" and "heater shield trim."
> 
> I think this makes sense, because for the most part, these kinds of assets would always appear right next to its prerequisite asset anyway.

Okay, this should be doable.

> 
> #### 2. Excluded Assets:
> For an "exclude" asset, once an asset that excludes it is enabled, we should list off what asset is blocking it.
> 
> For example, if you enable a specific cape, we can grab the cape from our asset selection and show in the category "This cannot be used with X" where X = the current selected asset.
> 
> Idea for steps:
> 
> 1. Choose a Dragonfly Wing Asset
> 2. Open the Capes Category
> 3. Where it says "no cape", instead show "Cape cannot be used with Dragonfly Wing"
> 
> These two things are optional, but I think they would make this feature easier to use.

This is doable too. However, I'm thinking we should hide the radio button next to "No Cape" as it will serve no purpose.

sheet_definitions/cape_solid.json
Gaurav0: > (Personally I would not restrict cape if wings or backpack are enabled, but I understand this is just an example.)

I think I'm just going to undo all my changes to sheet definitions before merging unless you let me know what changes you want.

sheet_definitions/shield_heater_revised_trim.json
Gaurav0: I don't think I can do one behavior if its next to and a different one if they are far apart.

sheet_definitions/cape_tattered.json
Gaurav0: > Okay, so, there's one weird issue here. The capes work fine once you disable backpack or wings. However, if you had the category open and the assets get filled out, it doesn't generate the previews.

I'll treat this one as a bug and fix before merging.

sheet_definitions/backpack_contents_ore.json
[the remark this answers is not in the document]
Gaurav0: > I really like how this works, though only a minor complaint, I would prefer if the category backpack cargo wouldn't even appear (to keep it consistent with, for example, body models).
> 
> The assets inside the category is perfectly fine, though, and I like that part. This isn't a blocker for the PR, just pointing out a preference. Perhaps its better this way anyway, but perhaps it might be confusing that there's a category with nothing in it...
> 
> Making the entire category appear when you enable a backpack might make it less confusing, but its hard to say.

It is. We have situations where all the subcategories under a category disappear, and the category is emtpy

Screenshot 2025-02-04 at 6 38 45 AM

The above occurs whenever body/sex is anything other than "male".

Gaurav0:
To be fixed in this PR before merging:

- [x] replace "No cape" with "cape is not allowed with wings"
- [x] regenerate previews after removing exclusionary asset
- [ ] revert sheet definition changes

To be discussed and fixed in a future PR:
- [ ] Remove asset from sheet that is excluded (but not automatically)

sheet_definitions/shield_heater_revised_trim.json
jrconway3: Right I wasn't advocating for that. I was just leaving comments as I reviewed each piece and eventually I came to what I believe is the better solution in the main review:

If an asset requires another tag, then it shouldn't appear until the prerequisite tag is enabled. These types of assets will usually appear next to each other anyway, because "required" mostly applies for things like trims.

jrconway3:
> This is doable too. However, I'm thinking we should hide the radio button next to "No Cape" as it will serve no purpose.

Right, I agree. That's what I meant. Replace the entire line (the radio button and "No Cape").


> There are categories hidden by body/sex type. Should we do the same there?

Yes, that is what I want if "required" is set on an asset. If an asset requires another asset, it should be hidden by default. I don't think this is a good idea for excludes, like I said, the best thing there is to say what other asset locks it down.


> revert sheet definition changes

Don't worry about this. Once the functionality is ideal, I'll make the changes I want to see here and open a PR into this one.


> Remove asset from sheet that is excluded (but not automatically)

The problem is you can already make the selection itself disappear by changing the prerequisite asset. In addition, this already happens for things like body type anyway. If an asset doesn't have a body type entry and you switch off it removes the asset.


One of the main reasons for implementing this functionality is to automatically remove a trim layer if we disable the asset that is used by said trim layer, to make the UI easier on the end user.

jrconway3 (approved):
I think there's still more than can be done, but this looks really good so far. Don't worry about the sheet definitions, we can open a new card to address those and I'll work on adding them in.

I still think we should remove assets if we remove the prerequisite. That's one of the big things I was suggesting in the original card. But your suggestion is to wait and review it later, so this should be fine for now.

PR #151, opened 2025-02-05
reduce size of index.html

Gaurav0:
If we could asuume that credit info such as licenses, notes, and authors are the same for all layers and variants rather than specific to every image file, we could have reduced the size of index.html by about 88%, from 105 MB to 12 MB

After implementing a fallback to the current method in cases where variants or layers have different licenses, notes, or authors, the size increased a bit back to about 15MB. That is still a 85% decrease.

```
$ git show master -- index.html | wc -c
 105354845
$ git show HEAD~1 -- index.html | wc -c
 12486407
$ cat index.html | wc -c
 14699804
```

### DIFF
--- a/scripts/generate_sources.js
+++ b/scripts/generate_sources.js
@@ -10,6 +10,8 @@ require("child_process").fork("scripts/zPositioning/parse_zpos.js");
 const es6DynamicTemplate = (templateString, templateVariables) =>
   templateString.replace(/\${(.*?)}/g, (_, g) => templateVariables[g]);
 
+const templateHTML = fs.readFileSync("scripts/template-general.html", "utf8");
+
 const licensesFound = [];
 function searchCredit(fileName, credits, origFileName) {
   if (credits.count <= 0) {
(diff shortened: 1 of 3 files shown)


jrconway3 (approved):
I'm not seeing anything particularly notable here.

I can confirm that the original file size of index.html is:
198 B

https://liberatedpixelcup.github.io/Universal-LPC-Spritesheet-Character-Generator/#?body=Body_color_light&head=Human_male_light

And the NEW file size here is:
97 B


So yeah its greatly reduced in size, though it still doesn't seem that big either way.

Gaurav0:
I mainly did this out of concern for our git repository size growing. Uncompressed file size on disk is what I measured.

However, if you are talking about the download size (after compression), you need to turn off the cache and make sure you are looking at the document, not the redirect.

Before:

image

After:

image

Yes, index.html is highly compressible, especially on modern servers using bzip2 rather than gz compression!

The above is on my fast cable modem. If I throttle down to "Fast 4G", It goes from about 6 s to 3 s.

PR #154, opened 2025-02-06
implement eyes for face option 1

Gaurav0:
This is for discussion #150 Option 1.

This is based on #151 , please review that PR first.

### DIFF
--- a/scripts/generate_sources.js
+++ b/scripts/generate_sources.js
@@ -2,7 +2,8 @@ const fs = require("fs");
 const readline = require("readline");
 
 const DEBUG = false; // change this to print debug log
-const onlyIfTemplate = true; // print debugging log only if there is a template
+const onlyIfTemplate = false; // print debugging log only if there is a template
+const onlyIfReplace = true; // print debugging log only if there is replace key
 
 require("child_process").fork("scripts/zPositioning/parse_zpos.js");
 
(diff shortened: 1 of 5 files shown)


jrconway3 (approved):
This is looking really good, nicely done!

jrconway3:
> This is looking really good, nicely done!

```
      "Closed_Eyes": "default",
      "Blush": "default",
      "Shame": "default"
```

Ah, I forgot about this. This is a bit of an issue, we would need some way to disable it on these entries. I guess I could just make a "none" category, though alternatively, I could set the faces here to appear above the eyes and otherwise not worry about it too much.

I think that's what I'll do, so this should be fine.

Gaurav0:
> 
> ```
>       "Closed_Eyes": "default",
>       "Blush": "default",
>       "Shame": "default"
> ```
> 
> Ah, I forgot about this. This is a bit of an issue, we would need some way to disable it on these entries. I guess I could just make a "none" category, though alternatively, I could set the faces here to appear above the eyes and otherwise not worry about it too much.
> 
> I think that's what I'll do, so this should be fine.

I think a none category would look better.

jrconway3 (approved):
This will work fine. I'll make any remaining fixes after merge, so don't worry about that. As long as the functionality is working so we can implement the toggle we're good and its looking good to me here.

PR #156, opened 2025-02-08
Add python image processor

kjenney:


jrconway3:
Could you start a discussion to explain what's going on here so we can determine if its something that should be added to the repo?

I'm not inherently against it but I'd rather have a discussion thread where multiple people can chime in.


As it stands, though, at the very least I am against adding all that documentation to the main README file. I am fine if you want to add that to its own README file and link it from the main one.


One the other hand... another strong possibility is to create your own repo, drop it into the LPC organization, and have this repo as a submodule. Then you won't have to also manage all the spritesheet files.

jrconway3:
Its all right. Thanks for the clarification.

Issue #161, opened 2025-02-10
Generate Sources Cannot Detect Conditional Paths for Credits


jrconway3:
When generating credits, if you use conditional paths then the credits fail to generate:
```
PS E:\code\assets\lpc\Universal-LPC-Spritesheet-Character-Generator> node scripts/generate_sources.js
missing credit after searching recursively filename: head/faces/${head}/neutral/light
missing credit after searching recursively filename: head/faces/${head}/closing/light
missing credit after searching recursively filename: head/faces/${head}/closed/light
missing credit after searching recursively filename: head/faces/${head}/look_r/light
missing credit after searching recursively filename: head/faces/${head}/look_l/light
missing credit after searching recursively filename: head/faces/${head}/eyeroll/light
missing credit after searching recursively filename: head/faces/${head}/shock/light
missing credit after searching recursively filename: head/faces/${head}/sad/light
missing credit after searching recursively filename: head/faces/${head}/happy/light
missing credit after searching recursively filename: head/faces/${head}/blush/light
missing credit after searching recursively filename: head/faces/${head}/shame/light
Updated z_positions.csv!
CSV Updated!
Found licenses: [ '0', '1', '2', '3', '4' ]
HTML Updated!
```

This is in relation to #160. In my comment at the bottom, I suggested implementing this:
> The other possibility is simply removing the template mechanic entirely from the faces and ONLY using replace_in_path, something like this:
```
  "tags": ["expression"],
  "required_tags": ["human"],
  "replace_in_path": {
    "head": {
      "Human_male": "adult",
      "Human_male_small": "adult",
      "Human_male_plump": "adult",
      "Human_male_gaunt": "adult",
      "Human_female": "adult",
      "Human_female_small": "adult",
      "Human_male_elderly": "elderly",
      "Human_female_elderly": "elderly",
      "Human_elderly_small": "elderly"
    }
  },
  "layer_1": {
    "zPos": 101,
    "male": "head/faces/${head}/anger/",
    "muscular": "head/faces/${head}/anger/",
    "female": "head/faces/${head}/anger/",
    "teen": "head/faces/${head}/anger/",
    "pregnant": "head/faces/${head}/anger/"
  },
```

[...]

Issue #160, opened 2025-02-09
Cannot Match Eyes to Faces By Type


jrconway3:
So in the eyes I have no ability to implement conditionals from the face asset conditionals, like so:
```
<!-- Male -->
<li>
	<span class="condensed">Male Expressions</span>
	<ul>
		div_sheet_face_neutral%head=male
		div_sheet_face_closing%head=male
		div_sheet_face_closed%head=male
		div_sheet_face_look_r%head=male
		div_sheet_face_look_l%head=male
		div_sheet_face_eyeroll%head=male
		div_sheet_face_shock%head=male
		div_sheet_face_angry%head=male
		div_sheet_face_sad%head=male
		div_sheet_face_happy%head=male
		div_sheet_face_blush%head=male
		div_sheet_face_shame%head=male
	</ul>
</li>
```

In the eyes, I only have access to the name of the asset:
```
  "replace_in_path": {
    "expression": {
      "none": "default",
      "Neutral": "neutral",
      "Closing_Eyes": "closing",
      "Closed_Eyes": "none",
      "Looking_Left": "look_l",
      "Looking_Right": "look_r",
      "Rolling_Eyes": "eyeroll",
      "Shock": "shock",
      "Angry": "anger",
      "Angry_Alt": "default",
      "Sad": "sad",
      "Sad_Alt": "sad2",
      "Happy": "none",
      "Happy_Alt": "none",
      "Blush": "none",
      "Shame": "shame"
    }
  },
```

This prevents me from being able to set up "elderly" eyes as well which have a different structure.

One possibility here is I could just make elderly eyes its own sheet. Its annoying, but not nearly as bad as having to choose between a dozen eye assets.

I considered the possibility of restricting the elderly eyes only to eldery human heads and then matching the heads to the eye swaps instead. However, that comes with its own complications:

I still can't restrict some assets conditionally, unless these values can be swapped in like so:
```
  "tags": ["expression"],
  "required_tags": ["human", "${head}"],
```

With `head` being this value:
```
div_sheet_face_shame%head=male
```

If it is, then I could toggle male and female to be adult and elderly to be elderly to have the correct conditionals:
```
      "Human_male": "adult",
      "Human_male_small": "adult",
      "Human_male_plump": "adult",
      "Human_male_gaunt": "adult",
      "Human_female": "adult",
      "Human_female_small": "adult",
      "Human_male_elderly": "elderly",
[...]

Issue #78, opened 2025-01-17
Hair Extensions Appear Below Long Ears


jrconway3:
Hair Extensions appear below long/furry ears. I tested this with Feathered Ears specifically.

This is a minor issue, just something I noticed while testing #71.

Not high priority, but shouldn't be hard to fix.

jrconway3:
Yeah this is happening on all ears except "Big" ears. Got a quick fix for this.

jrconway3:
Ah, so there's the problem. The ears should be above the hair, but below the hair extensions due to their positioning.

jrconway3:
Got a fix ready. I'll open a PR combining this and #153

jrconway3:
Whoops forgot to close this.

PR #162, opened 2025-02-10
Issue #153 Conditional Heads, Faces, and Eyes

### LINKED ISSUE #153: Add Conditional Assets for Facial Expressions
In the Face and Eyes pages, add conditional assets.

I still want a cleaner Eyes structure, but we can use this for now for the eyes.

jrconway3:
Added conditionals to heads, faces, and eyes, and fixed layering for ears (#78). Also fixed #153 and added fixes for #160 (temporarily).

For #160, this is a possible solution. Its the best situation to handle this for now. However, I would rather not have the faces restricted only to matching heads, but ideally available on any human heads.

#161 is a problem, though. I have a work around, but implementing more specific credits does not work correctly right now.


jrconway3:
I don't understand why the build failed. I can't see any issues with the code locally.

jrconway3:
I think it was a fluke. Just a random error that happened. I made a small fix to a single file and regenerated the source (I legit just noticed an issue with the credits on eye color). Anyway, when I pushed it generated fine. The differences were miniscule, so I just think at the time I pushed last time Nelify was going through an issue.

PR #172, opened 2025-03-04
Fix revised shoe assets

jrconway3:
I forgot to check the revised socks/shoes for cleaned up pixels it would seem. So I did that and also changed a few URL's to my  releases to shorten them.

Resolves jrconway3#73


jrconway3:
I'm working on cleaning up human heads as well (jrconway3#74) but I'll open a separate PR for that. Its only replacing the assets, not sheets, so it shouldn't hurt anything here.

sanderfrenken (approved):

PR #185, opened 2025-05-03
add Chinese readme file

ZhianLin:
add Chinese readme file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+[中文说明](./README_ZH.md)
+
  LPC Spritesheet Character Generator
  =============================================
 
--- a/README_ZH.md
+++ b/README_ZH.md
@@ -1,0 +1,122 @@
+# LPC 精灵表角色生成器  
+=============================================
+
+本生成器尝试整合迄今为止所有由[LPC](https://lpc.opengameart.org)创作的角色素材。
+
+[点击此处在线试用](https://liberatedpixelcup.github.io/Universal-LPC-Spritesheet-Character-Generator/)。
+
+Liberated Pixel Effort 是由多位优秀艺术家共同协作的成果，他们为该项目创作了大量精灵素材。  
+**若您希望在项目中使用LPC精灵素材，必须为所有贡献者署名**，具体方式请参见[下方署名要求](#许可协议与署名要求)。
+
+虽然本仓库主要聚焦角色精灵素材，但LPC还包含许多瓦片集和其他艺术作品。瓦片集合集可在[OpenGameArt.org](https://opengameart.org)获取。
+
+### 历史沿革
+
+Liberated Pixel Cup的概念由Bart Kelsey和Chris Webber提出，最初是[OpenGameArt.org](https://opengameart.org)上由Creative Commons、Mozilla和自由软件基金会赞助的比赛（注：这些组织与本生成器无关）。其目标是建立遵循统一[风格指南](https://lpc.opengameart.org/static/LPC-Style-Guide/build/index.html)的艺术素材库。
+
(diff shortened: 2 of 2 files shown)


jrconway3:
I unfortunately don't understand Chinese so I can't know if this is accurate or not.

I think my main recommendation would be to add a `lang` directory and forward alternate languages there so as to keep the index a bit cleaner.


I definitely don't approve of this change, though:
```
[中文说明](./README_ZH.md)

LPC Spritesheet Character Generator
=============================================
```

This shouldn't be above the main header. I would instead put in a subheader with "Languages" (or "Translations"), like this:
```
LPC Spritesheet Character Generator
=============================================

#### Translations
[Chinese / 中文说明](./lang/README_ZH.md)
```


As long as the translation is accurate, I don't have any problems adding alternate languages. But I would want to add them in a cleaner way.

README.md
```diff
@@ -1,3 +1,5 @@
+[中文说明](./README_ZH.md)
+
  LPC Spritesheet Character Generator
  =============================================
```
jrconway3, line 5: I don't like having this above the main header. It should be below it, something like this:
```
 LPC Spritesheet Character Generator
 =============================================

## Translations
- [中文说明 / Chinese](./lang/README_ZH.md)
```

README_ZH.md
```diff
@@ -0,0 +1,122 @@
+# LPC 精灵表角色生成器
```
jrconway3, line 1: I am fine with this as long as the translation is accurate, but I do think the file should be in a `lang` directory like so:
```
lang/README_ZH.md
```

Something like that.

README_ZH.md
```diff
@@ -0,0 +1,122 @@
...
+
+- https://pflat.itch.io/lpc-character-generator  
+- https://vitruvianstudio.github.io/
+
+### 工具集
+
+- [lpctools](https://github.com/bluecarrot16/lpctools)  操纵像素艺术精灵和tilesets的工具。
+- [安装指南](tools/LPCTOOLS.md)  
+- [使用lpctools重新编译完整精灵表](tools/REBUILD.md)  
+- [素材转换工具](tools/VITRUVIAN.md)
```
jrconway3, line 119: Unless you're going to translate these sub-readme files as well, I would probably remove these. There's not much of a point in linking to them unless they're also translated, and these readmes are somewhat niche anyway.

But this isn't a requirement or anything, just my suggestion.

jrconway3 (changes requested):
Left a few comments addressing my main concerns. I'm impressed someone went through the effort of doing a full translation. This is a bit new to this repo, though, so I do think we should get some feedback from others as well before going through with this.

That said, if anyone else thinks differently from me on these issues I have no problems with relenting on my personal suggestions. These are just my thoughts. Gaurav, Sanderfrenken, or Bluecarrot might have different idea.

Loufe:
A quick research showed the following demo repo as a respected standard for enabling multilingual README files (https://github.com/jonatasemidio/multilanguage-readme-pattern?tab=readme-ov-file). Personally I like the look and professional air of the buttons.

To your point, @jrconway3, I agree it's worth thinking hard on a structure first. A "documentation" directory with subdirectories for each document (in which any associated images and every copy in all languages can be found, aside from the default English readme) might work. What do you think?

jrconway3:
I guess that could work, but I'm still concerned about crowding up the index. I do however really like this part right here:

image


Have these icons right underneath the main header for languages. I think that's a really good structure.

Loufe:
Any thoughts on directory structure, @Gaurav0 ?

Gaurav0:
I agree with @jrconway3 that all documentation files besides English should be in a subdirectory. I would actually prefer an even more nested directory structure:

lang/zh/README.zh.md
lang/fr-ca/README.fr-ca.md

I do like the icons.

PR #191, opened 2025-05-26
Fix eyes color not working

sanderfrenken:
By selecting a default expression.


jrconway3:
This isn't the correct solution. I'll have a better fix shortly.

PR #192, opened 2025-05-27
* Resolved Default Eyes Broken, Fixed Issue With Spellcast Female Fac…

jrconway3:
- Resolved Default Eyes Being Broken (from #191)
- Fixed Issue With Spellcast Female Faces
- Fixed Neutral Eyes on Spellcast


jrconway3:
I tried fixing this as well:
https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/issues/161

But it was too convoluted, so I pushed it back for now.

sources/chargen.js
```diff
@@ -953,6 +958,7 @@ $(".exportSplitAnimations").click(async function() {
   function dynamicReplacements(itemToDraw) {
     const { fileName, name, parentName, variant } = itemToDraw;
     const el = document.getElementById(`${parentName}-${name}_${variant}`);
+    console.log('dynamic replacements');
```
sanderfrenken, line 961: IMO this log can be removed from here

sanderfrenken:
Nice fix! Thanks for looking into it :)

sanderfrenken (approved):

PR #198, opened 2025-07-04
fix some link formatting in README

Gaurav0:


jrconway3 (approved):
Not sure why this is required but okay, I don't see anything wrong with it.

PR #199, opened 2025-07-05
Issue jrconway3#34 Add longsleeve male shirts

jrconway3:


Gaurav0:
> [Renamed Shortsleeves Directory to shortsleeves2](https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/pull/199/commits/29d1030d789439c2b94f0f95192bf6667f65c557)

Does this break any imports?

Gaurav0 (dismissed):

jrconway3:
One last confirmation here:

Just out of curiosity I started laying over the "base" shirts" over these shirts. The sleeves are subtly different for shortsleeves, though the longsleeves are significantly different. For the Shortsleeves, I can probably get away with using my Shortsleeves over Run/Jump/Revised Combat/etc. (I'll keep the base asset sleeves the same and use those sleeves for idle, jump frame 1, and emote frames 1 and 2).

For the Longsleeves, however, I'll have to make significant changes to make them consistent. They still won't be as bad as making them completely from scratch, though.

PR #200, opened 2025-07-16
Add JS script to recompose universal sheets from individual animation…

sanderfrenken:
Add JS script to recompose universal sheets from individual animation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+_universal/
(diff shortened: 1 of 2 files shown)


jrconway3:
Looking it over, I don't really think this is a good idea. What this will do is greatly increase the number of files we have. There are other ways to compile sheets. I put in documentation to use LPC Tools to implement this functionality.

If you REALLY want to add this, I would highly recommend disabling adding these folders into commits using `.gitignore`. Personally, I would rename the directories from `universal` to `_universal` and then hide `_univeral` using `.gitignore`.

jrconway3 (approved):
I can't get magick working locally... but at the very least, I can confirm that when `_universal` directories are created, they aren't consider viable to be committed, so this should work.

PR #211, opened 2025-08-28
Fix collision of exported item animation filenames

IoriBranford:
Base item animation filenames upon spritesheet filenames, since item spritesheets can have identical parentName/name/variant (such as foreground and background sheets for the same item)

### DIFF
--- a/sources/chargen.js
+++ b/sources/chargen.js
@@ -792,8 +792,7 @@ $(".exportSplitAnimations").click(async function() {
 });
 
   const getItemFileName = (item) =>
-    `${item.zPos}`.padStart(3, '0')
-      + ` ${item.parentName} ${item.name} ${item.variant}.png`;
+    `${item.zPos}`.padStart(3, '0') + ` ${item.fileName.replace(/\//g, ' ')}`;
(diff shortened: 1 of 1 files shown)


Gaurav0:
This will likely break anyone currently relying on the existing filenames in our exports.

Is there a reason you need this?

IoriBranford:
Without this fix, if two layers have identical z position + parentName + name + variant, only one of the layers will be included in the split item export.

Although that shouldn't be the case anyway so if it's preferred to fix individual items like #212 as they come up, that's OK.

Gaurav0:
That makes sense. Let's do it.

Gaurav0 (approved):
@sanderfrenken I think this is a good idea even if it breaks some existing code. The split animations by item feature is new enough that there should not be a lot of code based on it. Also, it's easy enough to fix existing code if necessary. I'd like to hear your thoughts if you have some time.

sanderfrenken:
I am also in favor of this change, I think fixing this issue is worth breaking the potentially existing integrations relying on the previous naming conventions.

PR #212, opened 2025-08-28
Fix crossbow's background z position

IoriBranford:
Set it to -1 like the bow's background


jrconway3:
Weird, not sure how this got missed. Thanks for the fix!

I see it here:
image


And it looks good on the deploy preview.

jrconway3 (approved):

PR #215, opened 2025-09-04
Add the Chinese README file under the lang directory.

ZhianLin:
Add the Chinese README file under the lang directory.


jrconway3 (approved):
Apologies for the delay, I saw it but forgot to respond. This should be fine, We still need to add the link from the main one but we can do that ourselves.

Issue #168, opened 2025-02-20
Performance is significantly worse


Loufe:
One of the game designers on my team has a not-so-powerful laptop. He's said the newest version is nearly unusable because of how much it has slowed down in the last couple months. 

Are there any obvious culprits for this? Is the way the animations were cut responsible?

I'd be happy to try my hand at fixing something but I'm not familiar enough with the internals to know where to start.

jrconway3:
A lot of updates have been made, and one concern was performance. Bluecarrot16 had concerns about performance as well.

That being said...


I have been keeping an eye out on performance and I haven't seen any issues with it. So I don't really know what to say about this. Can you provide more information about what's happening?


A number of updates that could be tied to performance is splitting the assets into multiple sheets, conditional assets based on other assets, eye color for example is a big one where the eyes adjust based on other assets (such as faces).

Loufe:
Are there any obvious ways to measure performance speed? Like are there any timers printed to debug which could help profile the performance of difference aspects of the tool, to your knowledge?

Gaurav0:
@Loufe I use the performance tab of Chrome DevTools

PR #219, opened 2025-10-06
Add basic profiler

cdvv7788:
image

This adds a basic profiler and measurements thorough the code. This is related to https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/issues/168. If you cannot measure it you cannot fix it, so I decided to start with that.

Tested only on chrome based browsers (I use brave). In firefox it didn't work even before my changes (may be something with my local setup?)

### DIFF
--- a/sources/chargen.js
+++ b/sources/chargen.js
@@ -88,7 +88,7 @@ $(document).ready(function () {
   let didStartRenderAfterLoad = false;
 
   const canvas = $("#spritesheet").get(0);
-  const ctx = canvas.getContext("2d");
+  const ctx = canvas.getContext("2d", { willReadFrequently: true });
   const images = {};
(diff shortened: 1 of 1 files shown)


Gaurav0:
> In firefox it didn't work even before my changes (may be something with my local setup?)

What doesn't work in Firefox? I'd like to hear more about that. Can you open an issue?

cdvv7788:
> > In firefox it didn't work even before my changes (may be something with my local setup?)
> 
> 
> 
> What doesn't work in Firefox? I'd like to hear more about that. Can you open an issue?

I tried in my mobile and other pcs (ios tho) and it did work, I think it may be my windows computer. I will open an issue if I can reproduce it somewhere else.

PR #230, opened 2025-10-14
Paint in place for real

Gaurav0:
I am not super familiar (or familiar at all) with this WebGL stuff but basically it goes like this:

- Select one of the parts we are going to be recoloring dinamically (for now only the pieces that are tied to the body color as they seem to be using the same palette)
- Create a canvas with a webgl context (shared - expensive operation performed once)
- Initialize a recolor cache
- Download the base variant for the asset (light)
- Apply recolor in the webgl canvas.
- Copy to an output webgl canvas so we can reuse the shared one for the next drawing.
- Copy to a `2D` context canvas we have offscreen. This is then copied once again to the preview canvas, and is the one used for downloads and stuff.

With CPU pixel by pixel it takes 190ms-230ms. With GPU it takes around 90ms-130ms.
The improvements are there and seem to be working. I may not be using the optimal setup tho.


cdvv7788:
> ```
> 
> /**
> 
>  * Palette configuration - maps item category paths to palette files
> 
>  * Each palette config defines which items use that palette based on their metadata path
> 
>  */
> 
> const PALETTE_CONFIG = [
> 
> 	{
> 
> 		type: 'body',
> 
> 		file: 'tools/palettes/ulpc-body-palettes.json',
> 
> 		sourceVariant: 'light',
> 
> 		categories: [
> 
> 			['body', 'body'],           // body-body
> 
> 			['head', 'heads'],          // all heads
> 
> 			['head', 'ears'],           // ears
> 
> 			['head', 'nose'],           // noses
> 
> 			['head', 'head_wrinkles'],  // wrinkles
> 
> 			['head', 'face']            // facial expressions (happy, sad, angry, etc.)
> 
> 		]
> 
> 	}
> 
> 	// Future palette types will be added here:
> 
> 	// { type: 'hair', file: 'tools/palettes/ulpc-hair-palettes.json', ... }
> 
> 	// { type: 'cloth', file: 'tools/palettes/ulpc-cloth-palettes.json', ... }
> 
> 	// { type: 'eyes', file: 'tools/palettes/ulpc-eye-palettes.json', ... }
> 
> ];
> 
> ```
> 
> 
> 
> Can't say I like this setup. My preference would be to choose exactly which palettes are supported from the sheet_definition json file. Maybe we can do this for common standards, but personally I'd prefer just specifying the material type of palette.
> 
> 
> 
> Even if we don't go that far, I think we can do better than including this in the `palette-recolor.js` file. This should be capable of being its own json config file, so if nothing else we could include the palettes block as its own json in the sheet_definitions or even in `tools/palettes`.
> 
> 
> 
> 
> 
> In addition, we might also want the ability to specify two different palettes, This is NOT a huge concern though, I'm find having a separate layer. But that would make things way easier to handle certain things like eye colors on faces.
> 
> 
> 
> 
> 
> Still, I will work with this for now. I just think it can be done better. I don't want the average user to be expected to go into the `sources` to choose stuff like this, json files are way easier to use.

I think your approach directly in the json file is better. We will define the palette file, and remove the variants (already defined in the palettes anyway)? Or what do you have in mind?
This should be relatively easy to implement. Do you agree with both changes to the JSON definition file?

jrconway3:
> I think your approach directly in the json file is better. We will define the palette file, and remove the variants (already defined in the palettes anyway)? Or what do you have in mind? This should be relatively easy to implement. Do you agree with both changes to the JSON definition file?

I'm still theorycrafting some ideas here, but I know Vitruvian (https://vitruvianstudio.github.io/) has the ability to choose between multiple types of palettes:
image

As well setting two different palette colors for a single asset
image


This kind of flexibility isn't mandatory or anything, but I don't see why we can't do it with this setup. The more complicated one is probably stacking multiple types.


However, one issue is definitely in how the default colors work. Currently it chooses the "basic" color from our chosen palette. That will only work, however, if we pick the one correct palette that has the matching "basic" color.

Maybe we can optionally decide which file to pull palettes from and which palette is the default. I'm pretty sure Vitruvian already specifies exactly what the default color is.

cdvv7788:
If we had an standardized set of palettes, this would be trivial, but that would need remapping a lot of assets. I think that something like this could work:
```json
[
{"body": {//body_palette here},
 "cloth": {//cloth_palette here}}
]
```

then in the meta json for the object:
```json
...
"base_palette": "body",
"base_color": "light",
...
```

If all of the palettes have the same length, this would be trivial to match. We could just display every palette we have as an option.
I did saw palettes with only 3 colors, maybe they can be remapped to use a 8 colors palette even if not all of the colors are in the asset?

This can be done later. For now I will update it to:
- select the base variant in the json file
- set the palette name in the json file

It should be easy to extend later if we decide to go for it.

cdvv7788:
@jrconway3 this should be more aligned with what you said. We can iterate from here later. Please give it a check :) (functionally should be exactly the same, just refactored the configuration surface).

cdvv7788:
@jrconway3 are you ok with this going to the refactor branch then? (Can you approve if that is the case please?)

cdvv7788:
> I'm trying to get the full_rewrite branch in a place we can merge it. I'm worried merging this will make that more difficult.

This is where our differences lie. If this branch is going to be separated from the other for a while then it makes more sense. I saw this just as another addition.

> I don't see how you are blocked. I understand that you are excited, I am too. But sometimes things take a while. Reviewing and testing are important.

Clearly defining the intention and objectives is important too. Because of the point above I do understand the discrepancy in our POVs tho.

> @jrconway3 and I have other stuff going on. I have a full time job and other side projects. I've been devoting extra time to this project because I don't want your work to go to waste or to lose your energy. I made you a member so that you can help out in other ways.

Not my intention to make you feel pushed, I just wanted to merge this if there was not a clear blocker. I am not a fan of merging back and forth all over the place.


In any case, please let me know if you find bugs in this branch, I will take care of them. I don’t want to base any further work off this until it is merged because if we change anything here then we would have yet another version to merge in more than one place.

cdvv7788:
Current version is broken:

```
palette-recolor.js:16 Uncaught ReferenceError: DEBUG is not defined
    at palette-recolor.js:16:1
```

PR #233, opened 2025-10-14
restore feature pinch to zoom

Gaurav0:


cdvv7788 (approved):
My only nitpick here is that pinching allows arbitrary zooms, beyond predefined limits. I think it should be limited to an specific range. Not a blocking one.

jrconway3 (approved):
Looks good to me.

PR #239, opened 2025-10-16
restore deleted keys from sheet definitions

Gaurav0:
As a first step to restoring backward compatibility, I wrote a script to restore missing keys to sheet definitions, and applied it. This should help with getting eye color working, with making old urls work, and with importing json. Checking in the script in case it is ever needed again. Also, the script can be easily adapted to make any sort of bulk updates to sheet definitions.


Gaurav0:
I don't think we should use the type_name going forward, but find a more flexible alternative.

But for now, we need it to restore backwards compatibility. We just do a translation based on it.

PR #248, opened 2025-10-19
Fix credits type replacements

Gaurav0:
Reuses the template dynamic replacements to fix missing credits


cdvv7788:
@Gaurav0 please check if this does work.

Added a default for the expression, because eyes were not being rendered(they depend on the expression).
Changed the path checking for credits.

Gaurav0:
Thanks! LGTM, but I'd like @jrconway3 or even @bluecarrot16 to look this over.

Gaurav0:
@jrconway3 Using exact filename:

Screenshot 2025-10-20 at 8 36 49 AM

PR #249, opened 2025-10-19
Poc tests

cdvv7788:
@Gaurav0 @jrconway3 while we finish up the full_rewrite requirements to get to the finish line, I think adding tests would be time well spent. More than once already I have found us fixing stuff more than one time after a change. This requires npm, but ONLY for the tests, and they will be ran as part of the github action. That should gives us a stronger guarantee that our changes are not breaking anything.

If you are ok with this, I will add tests for all of the components in the branch. Once we have a decent coverage we could enforce tests with coverage, but that is a decision for another day.


cdvv7788:
@Gaurav0 does something like this work? Not very proud of this (at all) haha

PR #252, opened 2025-10-23
feat: restore Replace Mask (Pink) feature

Gaurav0:


jrconway3 (approved):
Okay this really annoyed me, but at least its clearer here.

I checked the original and the pink mask button is up here apparently:
image


I literally never saw that button before.

I actually noticed the button this time because its directly above the preview window.

Anyway, this appears to be working as expected, but IMO I feel like this should just be enabled by default or even be a checkbox that can be toggled on/off.

PR #265, opened 2025-10-31
Fix Body Color expands or collapses with Body

Gaurav0:
"Body Color" and "Body" have the same itemId ("body") but different paths so they are shown as separate categories in the tree ("body" vs "body-body") but they expand or collapse at the same time.

This is a quick and dirty fix for this bug.

@cdvv7788 may be able to suggest something better.


cdvv7788:
If this fixes it, let's go with it. I think adding tests will help us lock this stuff in place and cleaning up better later.

PR #267, opened 2025-11-10
Custom anims in zip

Gaurav0:
Fixes zip file creation for custom animations.

Complete:
- [x] Fix split by animation
- [x] Fix split by item
- [x] Fix split by animation/item
- [x] Add loading animations to appear while generating the zip files

TODO as separate PRs:
- [ ] Improve performance by:
  - [ ] Not reloading the images
  - [ ] compressing zip files in parallel

### DIFF
--- a/sources/components/download/Download.js
+++ b/sources/components/download/Download.js
@@ -66,8 +66,11 @@ export const Download = {
 					downloadFile(csvContent, "credits.csv", "text/csv");
 				}}, "Credits (CSV)"),
 				m("button.button.is-small.is-info", { onclick: exportSplitAnimations }, "ZIP: Split by animation"),
+				state.zipByAnimation.isRunning ? m("span.loading") : null,
 				m("button.button.is-small.is-info", { onclick: exportSplitItemSheets }, "ZIP: Split by item"),
+				state.zipByItem.isRunning ? m("span.loading") : null,
 				m("button.button.is-small.is-info", { onclick: exportSplitItemAnimations }, "ZIP: Split by animation and item"),
+				state.zipByAnimimationAndItem.isRunning ? m("span.loading") : null,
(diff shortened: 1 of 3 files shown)


jrconway3 (changes requested):
In split by item, I don't see "custom" and only "items", and it puts the sword slash here (EDIT: No, its not the sword slash, its walk, the slash is missing entirely):
140 longsword__longsword_ png

There is no oversize slash here like there should be.


Split by animation and split by animations and items both work:
slash_oversize
thrust_oversize
slash_reverse_oversize



However, split by animation is very slow, while split by animation and item is ridiculously slow. While it works, I don't think I can approve it due to just how slow it is. Also, there's no indicators either.

Its also not like I have many items, I only added the sword:
https://deploy-preview-267--liberatedpixelcup.netlify.app/#sex=male&body=Body_color_light&head=Human_male_light&expression=Neutral_light&weapon=Longsword_longsword


EDIT:
Correction, split by animation and items doesn't work correctly. In "custom" I ONLY got thrust_oversize, its missing slash_oversize and slash_reverse_oversize:
140 weapon_sword_longsword_

Gaurav0:
@jrconway3 Thanks for reviewing and testing!

jrconway3 (changes requested):
Loading icon works. Still a bit slow, but it seems faster than it was last time. Last time it was ludicrously slow, now its a lot more reasonable.

Split by animations:
image

It puts it in the custom folder, but otherwise it works.


Split by items and animations also has all the correct anims, but again in the custom folder:
image



Lastly, split by items:

....completely and totally broken. Here's what the body, head, and face look like:
010 body_color__light_ png
100 human_male__light_ png
101 neutral__light_ png

It doesn't even have the proper full sheet, let alone the expanded. I don't know what's going on here, but it is cropped very poorly.



The ONLY thing that works correctly is the weapons sheet:
140 longsword__longsword_ png

So close but not quite there.

PR #270, opened 2025-12-03
Some idle animations

ChristianPervoelz:
PR contains a bunch of Idle animations for dresses, armor, necklaces and charms.


jrconway3:
So as it turns out, I had the Idles for Kimono lying around. Not a big deal, though, its only one extra frame really. I never added them because at the time I finished them I hadn't started working on expanded frames yet.

Issue #273, opened 2026-01-02
Proposition - Zip: Split per sprite


JernejHabjan:
That would allow splitting per individual sprite for example:
```
root:
  custom
  standard
    walk
      up
        1.png
        2.png
        3.png
        4.png
      down
        ...
  credits

This would allow easier bundling using texturePacker to custom atlases with framework-specific json files.

What do you think?

PR #274, opened 2026-01-02
#273 Split per frame

JernejHabjan:
## Changes
- added `ZIP: individual files` button which downloads individual frames grouped in folders (up,down,left,right) which are then grouped in anims.
## Fixes
https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/issues/273


JernejHabjan:
@Gaurav0 merged. Rdy for review

jrconway3 (changes requested):
I approve pending you fix Gaurav's request. It appears to be working well and you haven't broken the existing functionality. (There is a bug there, but its already broken on master, its not  related to  your fix. I'll fix it later.)

PR #275, opened 2026-01-05
UI: Enhance category tree with tooltips and item compatibility

Dammy8611:
I noticed that some sprites do not support all animations and to check this you either have to select a particular animation or scroll the Spritesheet preview or scroll all the way up to the current selections and hover each item to see a tool tip of the license and animations which is very annoying. I just took the tool tip logic from the current selection section and added it to the tree item variant so you can just hover and see the supported animations and licenses before selecting, saving time


Gaurav0:
This is great. Accepting.

PR #278, opened 2026-01-10
Will read frequently

Gaurav0:
This steals commit https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/pull/274/commits/403f2a2cffe79b08ef133fd3e4d8f6ace8b735d2 in #274 from @jHabjanMXP so we can apply it now.


jrconway3:
Honestly not sure what I need to be testing here...

PR #282, opened 2026-01-13
Enhance category tree

Gaurav0:
Addresses my review comments on #275

### DIFF
--- a/sources/state/state.js
+++ b/sources/state/state.js
@@ -2,6 +2,7 @@
 import { LICENSE_CONFIG, ANIMATIONS } from "./constants.js";
 import { syncSelectionsToHash, loadSelectionsFromHash } from "./hash.js";
 import { renderCharacter } from "../canvas/renderer.js";
+import { customAnimationBase, customAnimations } from "../custom-animations.js";
 
 // Global state
 export const state = {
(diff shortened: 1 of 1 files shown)


jrconway3 (dismissed):
So this is intended to replace the old PR and not merge into the old PR then?

The changes made to the old PR look good, but there are some minor issues I have with the overall thing.


Namely, this:
image

So the Whip says "no animation data" in it, resulting in it showing as compatible for everything.

And furthermore, all "custom animations" get messed up when their main counterparts are removed. In theory if a custom animation is for slash it should say slash is compatibile, even if its an oversized animation.


This issue is only happening because I JUST had a PR merged to fix some zip export issues. I didn't want it to try loading the standard animation if it doesn't exist, so I removed the standard animations from the list if its using custom oversize animations.


The only question is; should we take customs into account, or should we be saying that if a custom is an oversize of a standard animation, that animation is compatible? If so, then we should bring back the standard animations and put them on the sheet definitions again to fix this dilemma.

EDIT: I approved it anyway because I don't see this as a blocker, more something to fix later. I just want to bring up a possible issue I saw.

Gaurav0:
> Ideally it would be nice if custom animations could be filtered, too, if not by the "base" animation (slash > slash_128 / slash_oversize), or just by having all custom animations being filterable as well.

This should be working now. I feel this was a blocker because an item filtered by animation was not (un)selectable. The methods needed already existed. Please review.

Screenshot 2026-01-19 at 5 41 40 PM

Issue #234, opened 2025-10-15
More Flexible Recolor Options


jrconway3:
I'd really like to improve our options for how recolors work! This is in response to the PR https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/pull/230

I'm opening this as a separate issue to work on later.

### Problem 1:
Our palettes currently are too open in their palette count:
https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/blob/master/tools/palettes/ulpc-body-palettes.json
https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/blob/master/tools/palettes/ulpc-cloth-metal-palettes.json
https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/blob/master/tools/palettes/ulpc-cloth-palettes.json
https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/blob/master/tools/palettes/ulpc-eye-palettes.json

#### Solution:
Create new json files based on these that restrict all palettes to only 6 variants. These will be the basis for the new palettes based on material type.

Eye Color MAY be an exception here.

### Problem 2:
I'd like the ability to have more flexibility in palette recolors, similar to Vitruvian Studios:
https://vitruvianstudio.github.io/

What I would like is something like this:
```
  "recolors": {
    "color_1": {
        "label": "Skintone",
        "base": [
            "palette": "body",
            "color": "light"
        },,
        "palettes": [
            "body",
            "lpcr"
        ]
    },
    "color_2": {
        "label": "Eye Color",
        "base": [
            "palette": "eyes",
            "color": "blue"
        },
        "palettes": [
            "eyes"
        ]
    }
  }
```

The `color_1` and `color_2` would be the plausible options to swap colors with, and this is structured similar to what you see with `layer_1` and `layer_2`.

There's multiple things happening here:

1. `color_1`, `color_2`, etc is for the purpose of letting us have primary and secondary colors on a single asset. This mostly applies to body color so we can change both the body color and the eye color at the same time.
2. Each type of color has a label to show the user what color is being changed. If "label" is not provided, just "Color" can be used by default.
3. "Base" tells what color to use as the default. You pick the palette to use as the default and also the color from that palette.
[...]

PR #285, opened 2026-01-20
WIP: Issue #234 Hypothetical JSON Config Setup

jrconway3:
Worked on implementing JSON config setup for more advanced recolors.

This is an evolution of my concepts from #234. I have not finished this and it will throw errors if you test, this is just a WIP draft as a proof of concept.

One idea I had is that maybe the palettes should also be generated in item metadata? I don't know if that's necessary or not but it may improve performance. I dropped all palette files into a directory called palette_definitions, similar to sheet_definitions. This directory contains the palettes from tools, but properly configured to only have 6 palette colors each (3 for eye colors).

I also configured all body (and eye) colors that should work for this, though my  theory is that assets which have multiple colors should just have a multi-color setup. I have this configured to have only one by default, but it will auto-detect if multiple palettes exist.

Multiple palettes (mostly for skintone + eyes, but perhaps other assets can use this later) are configured similar to layers.

Also, I renamed "variants" to "colors". I don't know what else to call them, so different "versions" of palettes are called palette variants. Maybe there's a better name for it, but its the best I could come up with for now. Our default palettes are called ULPC, but LPC Revised palettes are also fairly popular so I wanted to support them for users who want them.

### DIFF
--- a/item-metadata.js
+++ b/item-metadata.js
@@ -65,7 +65,7 @@ window.itemMetadata = {
     "preview_x_offset": 0,
     "preview_y_offset": 0,
     "matchBodyColor": false,
-    "recolors": null,
+    "recolors": [],
     "licenses": {
       "male": [
         "CC0"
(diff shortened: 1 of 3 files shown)


jrconway3:
Okay I fixed it so that it doesn't break before loading the page. The interface still needs updating but I'll try working on that tomorrow.

(It currently doesn't show any recolors from paint-in-place. I'll at least fix it to display the recolors the same as before, but I'd like to redesign the UI slightly so it shows palette icons next to the preview.)

PR #286, opened 2026-01-20
cleanup: don't generate old-version index.html in generate-sources

Gaurav0:
Another step in the cleanup from the old version of the site.

This removes the code to generate the full index.html from sources/source_index.html and sources/template-general.html. It also deletes sources/template-general.html

Note that we were already not writing out the file, but we are still doing most of the work of generating it.

The next step is to remove the need for sources/source_index.html entirely so we can stop maintaining it. We could either:
- create a smaller json file to replace it (easier to code)
- walk the sheet_definitions directory tree (easier to maintain)


jrconway3 (approved):
I'm not seeing any issues with this, seems to be generating and running fine.

Issue #287, opened 2026-01-20
Remove the need for sources/source_index.html


Gaurav0:
Remove the need for sources/source_index.html entirely so we can stop maintaining it. We could either:

- create a smaller json file to replace it (easier to code)
- walk the sheet_definitions directory tree (easier to maintain)

Followup to #286

PR #288, opened 2026-01-21
Issue #287 Removed source_index.html

jrconway3:
Implemented #287 by parsing the sheet_definitions directory to get all files for metadata, then deleted source_index.html.

I also deleted three sheet definitions that don't have assets yet (there may be more, but these ones actually threw errors.)

NOTE: This is an extension of #286, so it includes those changes as well.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,5 +1,8 @@
 root = true
 
+[*]
+end_of_line = lf
+
 [*.js]
 indent_style = space
 indent_size = 2


jrconway3:
```
(base) PS E:\code\assets\lpc\Universal-LPC-Spritesheet-Character-Generator> git add -A
warning: in the working copy of 'CREDITS.csv', LF will be replaced by CRLF the next time Git touches it
warning: in the working copy of 'item-metadata.js', LF will be replaced by CRLF the next time Git touches it
warning: in the working copy of 'scripts/zPositioning/z_positions.csv', LF will be replaced by CRLF the next time Git touches it
(base) PS E:\code\assets\lpc\Universal-LPC-Spritesheet-Character-Generator> git config --global core.autocrlf false
(base) PS E:\code\assets\lpc\Universal-LPC-Spritesheet-Character-Generator> git add -A
(base) PS E:\code\assets\lpc\Universal-LPC-Spritesheet-Character-Generator> git status
On branch feature/issue-287-delete-source-index
nothing to commit, working tree clean
```

So I did that, and I'm not getting the warning anymore, but it still didn't fix the issue.

Gaurav0:
Hmmm.

My theory is that git is expecting that crlf is being used, but node scripts/generate_sources.js is only producing a lf at line endings. The solution would probably be to use `os.EOL` in the script rather than plain `\n`.

jrconway3:
I'll try that.

PR #289, opened 2026-01-22
Issue #287 Remove Source Index by Restructuring Sheet Definitions to Generate the Category Tree Layout

### LINKED ISSUE #287: Remove the need for sources/source_index.html

jrconway3:
Resolves #287.

I've organized all files and created meta json files for all directories.

The meta data for categories will fall back to defaults if nothing is provided, If label exists, then the label will replace the default category label (which is the name of the current directory but capitalized). If a priority exists, every item in that category depth will be sorted by priority. Any items that don't have a priority will be after the ones that do have one.

I did do some minor restructuring in the process as well. Most everything is in the same place, but I moved Hair and Headwear to top level categories now. Facial Hair is also under hair (Eyebrows are next to Eyes though, just makes more sense there IMO). I also organized Heads into sub-categories, though these can be subject to change. I mostly wanted Human Heads to be easily accessible.

I also fixed providing "required" body type on categories. That was missing after the MythrilJS restructure. Even better, it works on top-level categories (child has Weapons and Tools disabled) something that our old source_index.html file never supported.

I have not removed the path variables yet. I am pretty confident in how this looks so far, but feel free to point out any potential areas to fix.

Some things were also commented out, but I think they should just be removed. With the current structure we don't need the templating mechanic anymore:
```
  /*const templateIndex = fileName.lastIndexOf("%");
  let searchFileName = fileName;
  let queryObj = null;
  if (templateIndex > -1) {
    searchFileName = searchFileName.substring(0, templateIndex);
    const query = fileName.substring(templateIndex + 1);
    queryObj = Object.fromEntries(new URLSearchParams(query));
    const replObj = Object.fromEntries(
      Object.keys(queryObj).map(key => [key, ""])
    );
    searchFileName = es6DynamicTemplate(searchFileName, replObj).replace(
      /_+/,
      "_"
    );
  }*/
  ```

There's only a few additional issues I'd like to see fixed eventually:
1. I wanted to be able to intermix categories with items, but the current structure doesn't support that. Maybe it won't matter once we finish paint in place.
2. I think animations filters aren't working right now on categories, but I didn't see where that was on the subitems either.


jrconway3:
Yeah its failing again. Is it possible that the actual problem is that its reading the directory listing differently from what windows is doing?

jrconway3:
So I made the comparison by going into the nelify preview, grabbing item-metadata.js, copying it locally, and comparing to my local file. I don't know if that works because I assume that the test is in its own isolated environment, but if that's accurate this is the ONLY difference I see between the two:
image



I even replaced `EOL` with `os.EOL` and also set the global requirement to be `ls`.

Gaurav0 (dismissed):

jrconway3:
Everything looks good to me, but I can't merge (without bypassing rules) because "someone other than Gaurav must approve". Should we just skip the requirement and push anyway?

Gaurav0:
> Everything looks good to me, but I can't merge (without bypassing rules) because "someone other than Gaurav must approve". Should we just skip the requirement and push anyway?

Just approve it and you can merge without bypassing rules.

PR #295, opened 2026-02-05
Add more tests 2

Gaurav0:


Gaurav0:
How are you running tests, and what browser are you using?

Locally using Chrome, Safari, Firefox on Mac all pass.

In CI using Chrome and Firefox on Ubuntu all pass.

I'm kinda guessing this might be a Windows specific issues based on the 'process is not defined' error.

Gaurav0:
Ah. Try running `npm install` and then `testem` from the command line with Chrome and Firefox installed.

Gaurav0:
Try disabling Brave's canvas anti fingerprinting technology:

https://github.com/piskelapp/piskel/wiki/About-canvas%E2%80%90based-browser-fingerprinting-and-Brave-browser

jrconway3:
Ah that makes sense, its something Brave does as a security feature. Yeah, disabling it on localhost fixes the issue. I'll have to keep that in mind if I ever download stuff in the future as well, though I don't actually download images from ULPC in my browser as I just pull it in git instead.

PR #303, opened 2026-02-10
* Issue #302 Fixed Body zPos, Fixed Generating z_positions.csv

### LINKED ISSUE #302: We can't see shoes anymore
Cf: https://liberatedpixelcup.github.io/Universal-LPC-Spritesheet-Character-Generator/#sex=male&body=Body_Color_light&head=Human_Male_light&expression=Neutral_light&shoes=Revised_Shoes_black

Image

jrconway3:
I don't know what the point of `z_positions.csv` is, but I fixed it. I didn't realize those were a separate check, so the file got emptied because those were reading only the root of sheet definitions.

Bigger issue is this resolves #302 by fixing the body zpos. I screwed up the body zpos because I accidentally edited the "zPos" when trying to edit the priority.

### DIFF
--- a/scripts/zPositioning/parse_zpos.js
+++ b/scripts/zPositioning/parse_zpos.js
@@ -17,7 +17,7 @@ const files = fs.readdirSync(SHEETS_DIR, {
   const depthB = pb.split(path.sep).length;
   if (depthA !== depthB) return depthA - depthB;
 
-  return pa < pb ? -1 : pa > pb ? 1 : 0;
+  return pa.localeCompare(pb, ["en"]);
 });
 
 files.forEach(file => {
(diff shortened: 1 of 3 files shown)


scripts/zPositioning/parse_zpos.js
```diff
@@ -6,10 +6,21 @@ var csvEntries = [];
...
-}).forEach(file => {
+}).sort((a, b) => {
+  const pa = path.join(a.path, a.name);
+  const pb = path.join(b.path, b.name);
+
+  const depthA = pa.split(path.sep).length;
+  const depthB = pb.split(path.sep).length;
+  if (depthA !== depthB) return depthA - depthB;
+
+  return pa.localeCompare(pb);
```
Gaurav0: if you use localeCompare, please specify the locale(s). Otherwise another user may have a different locale set on his system and sort differently.

PR #305, opened 2026-02-12
* Issue #290 Fixed Exporting Custom Animations With Split FG/BG Layers

### LINKED ISSUE #290: Export by Item and Animation Isn't Exporting Both Foreground and Background Layers on Customs
Not sure how I missed this, but FG/BG aren't both exporting on custom animations. I was certain this was working, but its not right now. I'll double check it layer, its not related to #273 or #274 and merging that shouldn't break anything.

jrconway3:
Resolves #290 

One theorycraft here is maybe I should just only provide the zPos override instead of specifying the layer:
> export function getItemFileName(itemId, variant, name, layerNum = 1, zOverride = null) {

But I didn't want to rewrite the old stuff so I left it as is. I can't use the layerNum with custom animations because the sort order for fully custom animations is totally different. I didn't even think to just manually specify the zPos prior.

### DIFF
--- a/tests/utils/fileName_spec.js
+++ b/tests/utils/fileName_spec.js
@@ -67,4 +67,14 @@ describe('getItemFileName', () => {
     const result = getItemFileName(1, 'variant1', 'body_male_light.png');
     expect(result).to.equal('050 body_male_light.png');
   });
+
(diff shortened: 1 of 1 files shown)


Gaurav0 (approved):
Comment is non-blocking.

PR #378, opened 2026-04-02
* Removed Clothing Recolor Sheets (only for Fixed Assets)

jrconway3:
All Longsleeve, Shortsleeve, and Sleeveless assets already supporting recolors have the extra files removed.

I looked into the possibility of finishing up the remaining shirt assets in these categories, but all of them need color reductions so that's not a viable option without extra work. I'll check Vitruvian and see if there's assets over there I can pull over, in which case I'll credit Napsio for the color reductions.


Gaurav0 (commented):
For some reason, I'm not able to approve this PR.

Gaurav0 (approved):

PR #380, opened 2026-04-02
Fix custom animations missing standard items in split item animation ZIP

### LINKED ISSUE #382: Custom animations lack standard items in item animation ZIP
When exporting ZIP split by item and animation, any custom animations no longer include items from the standard animations they are based on.
 
1. Start a new character
2. Add a Halberd
3. slash_oversize and thrust_oversize animations are added
4. Export ZIP split by item and animation
5. Open the ZIP file

The slash_oversize and thrust_oversize only have sheets for the halberd. It should also include the body, head, clothes etc from the standard slash and thrust.
image

IoriBranford:
Fixes #382


jrconway3:
What exactly is this for?

Gaurav0:
@IoriBranford It looks like you are trying to fix a bug. Please open an issue and explain in detail what you are trying to fix, and how to reproduce.

jrconway3 (changes requested):
Under split by items and animations I see 4 custom animations when I enable Longsword:
Image

Slash Oversize, Reverse Oversize, and Thrust Oversize all look fine:
Image

However, Slash 128 and Walk 128 are missing the clothes:
Image


I don't see any reason why these would be missing, these clothes are on all frames.


There's another issue as well I'm trying to confirm, but I suspect that's happening on default and is my fault, not yours.

I will share it anyway, here's how the standard animations look:
Image

And here's how the customs look:
Image


I don't see anything in the code here that would stop recolors from working on the custom sheets, but I have yet to confirm this is happening on default. I'll review that just  to confirm, but I can easily already see some issues in assets missing as is.

jrconway3 (approved):
I will approve this now, but I absolutely think that its way too slow. I don't think its any slower than before but it definitely needs to be sped up somehow.

Split by item and frame is incredibly fast. Split by item and animation takes a very long time.

Unless Gaurav objects, I am fine with performance and documentation being ne PR(s).


The assets look good, though:
Image

Image


This is for "Longsword Alt" which is a custom animation only asset.

Gaurav0:
Yes, we can do all the other stuff in separate PRs.

Issue #381, opened 2026-04-03
Argos still shows visual regression when site is identical due to preview animation


Gaurav0:
See https://app.argos-ci.com/liberated-pixel-cup/Universal-LPC-Spritesheet-Character-Generator/builds/29/288198428 from #380 which doesn't change anything about how the site looks.

Argos is still showing differences between master and a PR because the preview is animating and the timing is different. We should turn off this animation while grabbing the screenshots.

PR #386, opened 2026-04-03
Issue #381: Dont run preview animation in visual test

Gaurav0:
See #381 

I recommend using the "hide whitespace" option when viewing the diff, git's default algorithm gets confused by the indentation changes.

### DIFF
--- a/sources/canvas/preview-animation.js
+++ b/sources/canvas/preview-animation.js
@@ -64,11 +64,89 @@ export function setPreviewAnimation(animationName) {
 }
 
 /**
+ * Draw one preview frame for a given index into `animationFrames` (the cycle).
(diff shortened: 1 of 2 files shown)


jrconway3 (changes requested):
I'm not sure whether to approve or not...

First of all, I tested everything locally and it looks fine. I don't see any issues. Lint looks good, tests run well.


However, in Argos I do see this:
Image


Its very clear that the new screenshot isn't static. I have no idea what's supposed to trigger this, though, is it not going to take affect until the next run? Or should it take effect now? I'm guessing it should be taking effect now, in which case its very clearly not working.

jrconway3 (approved):
I made a mistake, the walk animation skips the idle frame.

PR #396, opened 2026-04-06
Testing on browsers cross platform

Gaurav0:
1. Better windows compatibility
2. Do not attempt to run Safari on Windows or Linux even if it is installed
3. Open Safari in a way that doesn't show a security prompt for opening a local file
4. Fewer first run and UI prompts on Chrome/Firefox

I will also attempt to upstream some of this to testem today if it works well.

Update: Opened https://github.com/testem/testem/pull/1982 for (3) and https://github.com/testem/testem/pull/1983 for (4)

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,8 +8,8 @@
     "validate-site-sources": "node scripts/generate_sources.js",
     "z-positions": "node scripts/zPositioning/parse_zpos.js",
     "z-positions:update": "node scripts/zPositioning/update_zpos.js",
-    "test": "testem ci",
-    "test:server": "testem",
+    "test": "node ./node_modules/testem/testem.js ci",
+    "test:server": "node ./node_modules/testem/testem.js",
     "test:visual": "playwright test",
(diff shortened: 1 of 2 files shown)


jrconway3 (approved):
So here's what I'm seeing:

`npm run test` works great, I got this result:

> tests 598
> pass  598
> skip  0
> todo  0
> fail  0
> ok


Now, I ran `npm run test:server` to test opening browser windows. Weird stuff happening here.

First, looking at Chrome...

The errors are 297/299, with the zip files timing out:

Image


I don't know if there's a real way to solve this or not, just commenting on it. I'm still considering this a win.


The Firefox issue is far less important, just something weird to point out:

Image


The tests in firefox work perfectly fine. The only issue is that it created 14 "Problem loading page" tabs, which contain the following content:

Image


Each one has a different address in the address bar, so they seem to be something notable its trying to load and failing. It didn't break the tests or anything, so its just weird.

PR #404, opened 2026-04-09
* Fixed Zip Export Not Including Recolors in Custom Standalone Assets

### LINKED ISSUE #394: Some items don't respect color selections in ZIP by animation and item
In custom animations, items copied from standard animations (body, head, clothes etc) show with their selected colors on the canvas, but default colors in the ZIP by animation and item files.

1. Start new character
2. Give him a body color other than Light
3. Give him a large weapon that adds custom animations (e.g. long spear)
4. Download ZIP by animation and item
5. Open any custom animation in the zip

Other ZIPs don't have the issue, though ZIP by item has something similar to #382 where the custom animations don't have those items.

Image

jrconway3:
resolves #394 

The issue is that the new assets weren't configured correctly to use the recolor script.

I also added tests for it, too, with help from Copilot of course.

### DIFF
--- a/sources/state/zip.js
+++ b/sources/state/zip.js
@@ -12,6 +12,7 @@ import {
 import { getMultiRecolors } from "./palettes.js";
 import { getItemFileName } from "../utils/fileName.js";
 import { loadImage } from "../canvas/load-image.js";
+import { getImageToDraw } from "../canvas/palette-recolor.js";
 import { customAnimations, customAnimationSize } from "../custom-animations.js";
 import { getSortedLayersWithCustomFallback } from "./meta.js";
 import { canvasToBlob, image2canvas } from "../canvas/canvas-utils.js";
@@ -288,6 +289,7 @@ export const exportSplitItemSheets = async (deps = {}) => {
  * @param {typeof renderSingleItemAnimation} [deps.renderSingleItemAnimation]
  * @param {typeof loadImage} [deps.loadImage]
  * @param {typeof addStandardAnimationToZipCustomFolder} [deps.addStandardAnimationToZipCustomFolder]
+ * @param {typeof getImageToDraw} [deps.getImageToDraw]
  */
 export const exportSplitItemAnimations = async (deps = {}) => {
   const baseAddAnimationToZipFolder =
@@ -316,6 +318,7 @@ export const exportSplitItemAnimations = async (deps = {}) => {
   const renderSingleItemAnimationFn =
     deps.renderSingleItemAnimation ?? renderSingleItemAnimation;
   const loadImageFn = deps.loadImage ?? loadImage;
+  const getImageToDrawFn = deps.getImageToDraw ?? getImageToDraw;
 
   if (!guardZipExportEnvironment()) return;
 
@@ -431,6 +434,7 @@ export const exportSplitItemAnimations = async (deps = {}) => {
             `Exporting item ${itemFileName} for custom animation ${customAnimName}`,
           );
           let img;
+          let imageToDraw;
           let imgCanvas;
           await profiler.phase(
             "render_imageLoadDecode_customItemSprite",
@@ -442,7 +446,12 @@ export const exportSplitItemAnimations = async (deps = {}) => {
           await profiler.phase(
             "render_composite_customItemSprite",
             async () => {
-              imgCanvas = image2canvas(img);
+              imageToDraw = await getImageToDrawFn(
+                img,
+                layer.itemId,
+                layer.recolors,
+              );
+              imgCanvas = image2canvas(imageToDraw);
             },
           );
           if (!imgCanvas) continue;
--- a/sources/state/zip.js
+++ b/sources/state/zip.js
@@ -15,7 +15,7 @@ import { loadImage } from "../canvas/load-image.js";
 import { getImageToDraw } from "../canvas/palette-recolor.js";
 import { customAnimations, customAnimationSize } from "../custom-animations.js";
 import { getSortedLayersWithCustomFallback } from "./meta.js";
-import { canvasToBlob, image2canvas } from "../canvas/canvas-utils.js";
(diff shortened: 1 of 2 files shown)


tests/state/zip_spec.js
```diff
@@ -561,6 +561,21 @@ describe("state/zip.js", () => {
...
+      const c = document.createElement("canvas");
+      c.width = width;
+      c.height = height;
+      const ctx = c.getContext("2d");
+      ctx.fillStyle = `rgb(${r}, ${g}, ${b})`;
+      ctx.fillRect(0, 0, width, height);
+      return c;
+    }
+
+    function readTopLeftRgb(canvas) {
```
Gaurav0, line 574: Unit test this function. Make sure it handles edge cases (null or undefined passed as argument, non-canvas passed, blank or transparent canvas).

tests/state/zip_spec.js
```diff
@@ -561,6 +561,21 @@ describe("state/zip.js", () => {
       return c;
     }
 
+    function solidColorCanvas(r, g, b, width = 8, height = 8) {
```
Gaurav0, line 564: Unit test this function.

Gaurav0 (changes requested):

tests/state/zip_spec.js
```diff
@@ -561,6 +561,87 @@ describe("state/zip.js", () => {
...
+      const ctx = canvas.getContext("2d");
+      if (!ctx) {
+        throw new Error("Failed to get 2d context from canvas");
+      }
+
+      const data = ctx.getImageData(0, 0, 1, 1).data;
+      return [data[0], data[1], data[2]];
+    }
+
+    describe("helper functions", () => {
```
cdvv7788, line 588: Are we testing a test function? This would be ok if the helper was in the public surface, but it is here just to help us tests the actual code. Feels redundant.

sources/state/zip.js
```diff
@@ -442,7 +446,15 @@ export const exportSplitItemAnimations = async (deps = {}) => {
...
             "render_composite_customItemSprite",
             async () => {
-              imgCanvas = image2canvas(img);
+              imageToDraw = await getImageToDrawFn(
+                img,
+                layer.itemId,
+                layer.recolors,
+              );
+              imgCanvas =
+                imageToDraw instanceof HTMLCanvasElement
```
cdvv7788: This is a code smell, a function that returns 2 possible types. Maybe add a //TODO so we take care of it someday? ™️

cdvv7788 (commented):
Left a couple questions

tests/state/zip_spec.js
Gaurav0: If it is just a test function, it should have been only in the test code, not in sources/zip.js

sources/state/zip.js
Gaurav0: Yes, it looks like (not sure) we are returning a canvas in real code, and an image in tests.

tests/state/zip_spec.js
cdvv7788: `solidColorCanvas creates an 8x8 canvas by default` <- solidColorCanvas is defined in this same file at line 564

sources/state/zip.js
cdvv7788: Need to verify. If that is the case, we shouldn't muddy the implementation because of the tests, just muddy the tests. The function called does return canvas|image tho, and it is not touched in this PR, so that may not be the case.

tests/state/zip_spec.js
jrconway3: I've been going back and forth on this confused. At first I thought I made a mistake and allowed Copilot to put a test function into production, now I realize Gaurav is the one who made a mistake here. This is zip_spec and the test functions are in this file.

I can remove the test coverage for these methods if need-be, I only added them because Gaurav asked for them.

sources/state/zip.js
jrconway3: I'm checking if there's an easy way to just force getImageToDraw to always return a canvas no matter what. I don't see why we'd need to keep the image/canvas split.

jrconway3: Yeah so we don't need this at all, I just ran a test without it. Copilot thought it needed it because there are instances in zip.js where a canvas is used to split apart assets, but not in this instance.

And no, it wasn't because done for the sake of tests either, tests still work without it.

jrconway3: That being said, I do think getImageToDraw could just return a canvas every time. I don't see anything in there that requires us to return an image or canvas. But that's out of scope for this PR.

tests/state/zip_spec.js
jrconway3: @cdvv7788 Should I remove these? Again, I only added them because Gaurav explicitly asked for them. I was a bit confused, too, but I figured there was some nuance to it.

jrconway3: I don't think the tests are really hurting anything. :P

Gaurav0: You're right, my mistake.

Anyway, LGTM unless @cdvv7788 really feels we should remove the extra unit tests.

Gaurav0 (approved):

tests/state/zip_spec.js
cdvv7788: Is this a function called from the app or just from the tests? The latter is an anti-pattern because tests to test tests are...weird. I am just confused because of the file the function lives in.

jrconway3: Its only in tests and its only used in tests. Gaurav mistakenly thought it was in zip.js rather than zip_spec.js so he asked me to make a test for it.

I can easily remove the test but it works now and really isn't that important. It won't be a big deal whether its removed or we keep the tests.

cdvv7788: Not a huge deal, it is not a blocker :)

cdvv7788 (approved):

PR #405, opened 2026-04-09
* Implemented Skipping Animations in Credits if No Export is Enabled

### LINKED ISSUE #398: Credits Generates Non Animations as Filenames
> These new credits entries reference `legs/armour/plate/*/watering.png`, but there are no corresponding spritesheets in the repository. Either add the missing `watering.png` files or drop the `watering` credits entries (and the `watering` animation in metadata) to avoid pointing to non-existent assets.
> 
> _Originally posted by @Copilot in https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/pull/397#discussion_r3038300648_

I believe I added a no export flag for animations that don't have their own dedicated sheet, so I should be able to use that to filter out animation file references from the credits, too.

jrconway3:
resolves #398 

This skips animations that aren't standalone animations. If noExport is enabled in constants, the asset will be skipped when adding it in credits.


Gaurav0 (changes requested):
Agree with copilot. This will work in node 22.12+ only.

jrconway3:
> Agree with copilot. This will work in node 22.12+ only.

What's the alternative? I can't find anything that actually will work without duplicating the constants.js file.

Gaurav0:
Copilot gave you 3 alternatives. 1. Convert generate-sources.js to esm 2. Use a dynamic import. 3. Make constants.js not esm.

By esm I mean using import/export rather than require / module.exports (cjs). Before node 22.12 esm was backward compatible but not forward compatible. By which I mean an esm module could import a cjs module but a cjs module could not require an esm module.

A fourth alternative: Add an engines field to package.json making explicit that node 22.12+ is required and a line to CONTRIBUTING.md as well that says node 22.12+ is required.

My personal preference is (1). (4) is good also and it is easiest (plus we have to use 22 in ci anyway). (2) is ok as well, I don't really like (3), I'd prefer to move everything to esm (the new way) eventually.

jrconway3:
> Member

I tried the methods (except converting to ESM) and they didn't work. I didn't think converting to ESM in this context was a viable option so I basically ruled that one out completely. That one would work, but so far there hasn't appeared to be any interest/desire into converting this specific file into ESM so I didn't think it was something I should be looking into.

If you are fine with me converting it, though, then I will.


I tried to load via import + async but then running the script itself threw a fit. It worked, but threw this error:

```
(node:60424) [MODULE_TYPELESS_PACKAGE_JSON] Warning: Module type of file:///E:/code/assets/lpc/Universal-LPC-Spritesheet-Character-Generator/sources/state/constants.js is not specified and it doesn't parse as CommonJS.
Reparsing as ES module because module syntax was detected. This incurs a performance overhead.
To eliminate this warning, add "type": "module" to E:\code\assets\lpc\Universal-LPC-Spritesheet-Character-Generator\package.json.
(Use node --trace-warnings ... to show where the warning was created)
```

After going back and revisiting it, somehow I got this method to work correctly. Copilot created a duplicate file last time but this time it just dropped a package.json file in there to make it work.


However, I'll see if there's an easy way to just convert this to an esm file.

jrconway3:
I have pushed what I have for now. I'll look into converting generate_sources to .mjs _and_ adding test cases in a separate PR.

Gaurav0 (approved):

PR #406, opened 2026-04-10
Update master argos more pages longer views

Gaurav0:
The includes:

- more argos viewport sizes (long ones)
- another 2 "pages"
  1. where the modal is shown
  2. where search is done and everything is expanded
- scripts to dump and diff computed styles between 2 branches
  - These are basically being used to automate the process of figuring out what needs to be fixed to fix argos on the update_bulma branch.

This adds 20 new screenshots to argos but the original 4 are unchanged. Argos still shows an x, but says "20 added".


cdvv7788 (approved):

jrconway3 (approved):
The Argos pages look good and tests come back fine, so I think this should be good. I don't see anything notable with the code or anything.

PR #417, opened 2026-04-14
Meet open source sponsorship requirements for Netlify and Argos

Gaurav0:
This will not pass argos, of course. Deliberate change to site. See netlify preview.

Argos free trial will end today.

We're already sponsored by Netlify but I figure we should give Netlify and Argos equal billing.


cdvv7788:
image

Visual change only, both links work.

cdvv7788 (approved):

jrconway3 (approved):
I looked at it this morning and didn't approve because of the dark mode error. But since its not caused by this PR it should be fine.

Image

PR #421, opened 2026-04-15
Fix fixture builder issue 382

Gaurav0:
The script `npm run fixture:issue-382` wasn't working on master.


jrconway3 (approved):
It generates on its own and tests still work, but to fully verify this I created a branch off my other branch and this works.

PR #422, opened 2026-04-16
Revert to testem

Gaurav0:
Switches to vite but not from testem to vitest-browser


jrconway3 (dismissed):
This works fine and I can even get rsync to work, though it is pretty slow. Took over 30 minutes to run the rsync.

jrconway3 (approved):
Far as I can tell everything is working a lot smoother now. Tests actually finish and `npm run build` doesn't take forever. Tested `serve:open` and it looks pretty good, too.

jrconway3:
For some reason its now detecting some minor positioning changes? I don't see anything wrong with it, but keep it in mind.

PR #429, opened 2026-04-19
#416: make running generate sourcesjs part of build process and do not commit item metadatajs

### LINKED ISSUE #416: Make running generate-sources.js part of build process and do not commit item-metadata.js
We don't need this huge file in the repo once we switch over to vite, as we are not running the site directly from source.

First proposed in #413 but decided to split this out.

Gaurav0:
For #416 

Note that the plan generated was mentions vitest, we have reverted to testem.


jrconway3 (approved):
```
npm run build
npm run test
npm run test:server
```

Are all working pretty good. I even pulled an old branch I was working on and ran a test on that one (merged with this branch) and it worked great, too.

PR #432, opened 2026-04-20
Do not pretty print item-metadata.js in production

Gaurav0:
#### File Generation

The runtime UI loads **`dist/item-metadata.js`**, which Vite generates from the sheet JSON under **`sheet_definitions/`** (and related inputs) when you run **`npm run dev`** or **`npm run build`**. Do not edit that file by hand.

The JSON embedded in each `const` assignment is **pretty-printed** when Vite runs in development (**`npm run dev`**) and **compact** (no indentation inside those objects) when Vite runs a production build (**`npm run build`**). Inspecting **`dist/item-metadata.js`** after a dev run is easier; CI and release artifacts use the compact form.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ computed-style-diff-output/
 /dist/
 **/__screenshots__/
 /.vitest-attachments/
+.tool-versions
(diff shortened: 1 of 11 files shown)


jrconway3 (approved):
Had a weird thing happen where I generated it, it minified properly... then I ran dev and it STAYED minified for some reason.

I deleted the file, re-built, and it worked...

But then I tried it again after I re-built:
- Ran build (changed the existing "prettier" to minified)
- Ran dev (changed the existing minified to "prettier")

cdvv7788 (approved):

PR #436, opened 2026-04-20
Fix #435: Align directions in zip by frame

### LINKED ISSUE #435: Zip by animation and frame: left and down folders swapped.
See discussion #434

Gaurav0:
Correct LPC row order is up, left, down, right.


jrconway3 (approved):
Zip is correct and tests work, looks good.

PR #443, opened 2026-04-23
refactor: introduce TypeScript; migrate mask and constants to .ts

cdvv7788:
First step of a gradual TypeScript migration. Pure tooling + two pure-leaf file conversions to establish the pattern. No runtime behavior change; build output and test counts unchanged.

  ## Tooling
  - Add tsconfig.json (allowJs, checkJs:false, strict, moduleResolution bundler, noEmit, jsx off — Mithril uses hyperscript).
  - Add `type-check` npm script (`tsc --noEmit`).
  - Install `typescript` and `@types/node` as devDependencies.
  - Set `name` / `version` / `private` in package.json so the `package-lock.json` root name stays stable across worktrees (was flipping between "master" and the directory name).

  ## Conversions
  - `sources/canvas/mask.js` → `mask.ts` with explicit types (`HTMLCanvasElement`, `CanvasRenderingContext2D`). Removed the runtime null-guard — the type contract now enforces non-null, and the test case that exercised the guard is removed. This may leave us in a small void until we complete the migration, but cleaning up is part of this refactor. Once we type everything, defensive guards should no longer be necessary.
  - `sources/state/constants.js` → `constants.ts`. No annotations — inference suffices for pure data. Explicit types can be added later when a consumer benefits.

  ## Import strategy
  All importers of the two converted files now use explicit `.ts` specifiers (26 updates across sources/, scripts/, tests/). No loader required: Vite resolves `.ts` natively; Node 22.6+ strips types natively. Going forward, converting a file means renaming it and updating its importers in the same PR.

  ## Verification
  - `npm run type-check`: clean
  - `npm run lint`: clean
  - `npm run test:node`: 77/77 pass on plain node
  - `npm run build`: clean, output byte-shape unchanged
  - `npm run validate-site-sources`: runs on plain node, loads


Gaurav0 (changes requested):
1. Type check in CI
2. Update eslint config so typescript files are linted
3. Enable strictest ts compiler settings or document why not

Gaurav0 (approved):

PR #449, opened 2026-04-25
Use ESM everywhere except testem config file

Gaurav0:
- Set type: module; remove per-folder package.json that only set type
- Convert remaining Node scripts to ESM; rename testem.js to testem.cjs
- Rename former .mjs entrypoints/imports to .js; update Vite, tests, docs, and CI
- Default-export debug util for generateSources; ignore testem.cjs in ESLint
- Update documentation
- Recommend Node 24 as used in CI
- Add comment about testem.cjs vs testem.js 
- Require m from "mithril" to be imported: fixes errors in production 'm is not defined'

~~Note: argos failures are purely due to file renames.~~ fixed.


jrconway3 (dismissed):
Tests work, even fixing the tests that were creating during conversion to typescript that I fixed and pushed in a separate branch. So I guess I don't need that one anymore.

The only reason I didn't do this before is because there was too many commonJS stuff. But yeah I figured we'd eventually convert. Looks good so far. I don't see any obvious errors.

There is a sass deprecation I found, though:
```
Deprecation Warning [if-function]: The Sass if() syntax is deprecated in favor of the modern CSS syntax.

Suggestion: if(sass($right): "left"; else: "right")

More info: https://sass-lang.com/d/if-function

    ╷
413 │   $opposite: if($right, "left", "right");
    │              ^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ╵
    ..\..\..\bulma\sass\utilities\mixins.scss 413:14    @use
    ..\..\..\bulma\sass\base\skeleton.scss 3:1          @use
    ..\..\..\..\sources\styles\bulma-deferred.scss 5:1  @use
    ..\..\..\..\sources\styles\deferred-entry.scss 3:1  root stylesheet

Warning: 2 repetitive deprecation warnings omitted.
Run in verbose mode to see all warnings.
```


Recolors and exports still work, so I don't really see anything majorly wrong here.

PR #450, opened 2026-04-25
fix (sass): quiet deps flag

Gaurav0:
This flag quiets sass warnings from our dependencies, but not from our own code.

This silences the sass deprecation warnings you see from bulma.


cdvv7788 (approved):
Finally one I can read entirely!

PR #451, opened 2026-04-25
Refactor source generation: split entry points, Vite skip, validate via concurrently

Gaurav0:
- Add writeCredits to generateSources; extract writeZPositionsFromSheetsSync and
  thin generate_credits.js / parse_zpos entry points
- validate-site-sources runs generate_credits and parse_zpos in parallel via
  concurrently; CI uses npm ci and the same script
- Vite item-metadata: fingerprint sheet + palette trees (.cache/); skip when
  unchanged and dist metadata exists; otherwise refresh CREDITS, z_positions,
  and dist; VITE_REGENERATE_SOURCES=1 forces full run; dev watch reruns full pipeline
- Direct execution of generate_sources.js only prints to use validate-site-sources
  and exits 1 (library export for Vite/tests only)
- Update CONTRIBUTING, README, PALETTE_RECOLOR_GUIDE, debug.js; adjust tests

Made with [Cursor](https://cursor.com)


jrconway3 (approved):
Seems to be working well. I saw a weird issue where the styles were totally different when I ran my local instance:
Image


But it works fine on `npm run dev`:
image


And `npm run build` / `npm run dev` works just fine as well.

PR #455, opened 2026-04-27
Typed catalog API: collapse to single module + Result<T, LoadError> getters

cdvv7788:
## Summary

  Collapse the catalog's dual `getX(): T | undefined` + `isXReady(): boolean` API into a single `Result<T, LoadError>`-returning surface, and merge `catalog.js` (untyped state) + scattered local-cast types into one `sources/state/catalog.ts`.
  Behaviorally identical to `master` — internal cleanup, no UX change.

  ## API shift

  ```mermaid
  flowchart LR
    subgraph Before
      direction TB
      P["isLiteReady() : boolean"]
      G["getItemLite(id) : ItemLite | undefined"]
      P -.->|caller checks separately| G
      G --> U{undefined?}
      U -->|yes| Ambiguous["loading? not-found?<br/>(callers can't tell)"]
      U -->|no| Ok1["use item"]
    end
    subgraph After
      direction TB
      R["getItemLite(id) : Result<ItemLite, LoadError>"]
      R --> K{r.isOk?}
      K -->|yes| Ok2["use r.value : ItemLite"]
      K -->|no| Disc{r.error.kind}
      Disc -->|loading| Loading["chunk pending"]
      Disc -->|not-found| Missing["id absent"]
    end
  ```

  ## UI gating: `ResultBoundary`

  ```mermaid
  flowchart TD
    C[m ResultBoundary read view renderError]
    C --> Read["read returns Result"]
    Read -->|isOk| V[view value renders subtree]
    Read -->|isErr| E[renderError loading or not-found]
    V --> DOM
    E --> DOM
  ```

  Multi-chunk gates compose: `read: () => Result.combine([getItemMerged(id), getPaletteMetadata(), chunkReady('layers')])`. Components that only need a boolean (filters, ZIP buttons) call `isLiteReady()` directly — no boundary.

  ## What types buy us

  - **Discriminated `LoadError`**: `loading` vs `not-found` no longer conflated — visible in the type, branched in the consumer.
  - **Compiler-enforced narrowing**: can't read `.value` without `.isOk()`; can't read `.error` without `.isErr()`.
  - **One source of truth for shapes**: `ItemLite`, `ItemMerged`, `Credit`, `PaletteRecolor`, `LayerEntry`, `MetadataIndexes`, etc. — defined once. Audited against `dist/item-metadata.js`: 655/655 items have `name`, `type_name`, `required`,
  `animations`, `recolors`, `matchBodyColor`, `variants`, `path`, so those are required (not `?`).
  - **Zero `as ItemLiteShape | undefined` casts** remaining in `sources/canvas/`, `sources/utils/`, `sources/components/`. Deleted 4 duplicated local cast types.

  ## Still ironing out

  | Surface | Status | Path forward |
  |---|---|---|
  | Module-level singleton state (`let indexStage = ...`) | Kept; needed `resetCatalogForTests` | `createCatalog()` factory + DI (TODO in source); per-test instances |
  | Most consumers use `.unwrapOr(null)` | Throws away the discriminator | Convert consumer files to `.ts`; selectively branch on `r.error.kind` where loading vs not-found matters |
  | `getItemMerged` returns Ok with empty layers/credits if those chunks aren't ready | Intentional (preserves UX of item-names-before-credits rendering) | Documented at the function; revisit if/when we want strict semantics |
  | `load-failed` variant defined but unreachable | Dynamic-import failures still crash | Wire from `install-item-metadata.js` when we want recovery instead of crash |
  | Test files remain `.js` | Type narrowing not enforced in tests | Convert `_spec.js` → `_spec.ts` as a follow-up |
  | `neverthrow` dependency (~2KB gz) | Used: `ok`, `err`, `combine`, `.unwrapOr`, `.match`, `.isOk` | Could be replaced with ~75-line in-house Result if we ever want zero deps |
  ## Test plan


sources/state/catalog.ts
```diff
@@ -0,0 +1,541 @@
...
+  const lite = itemLiteStore?.[id];
+  if (!lite) return err(notFound(id));
+  const layers = layersStage.resolved ? (itemLayersStore?.[id] ?? {}) : {};
+  const credits = creditsStage.resolved ? (itemCreditsStore?.[id] ?? []) : [];
+  return ok({ ...lite, layers, credits });
+}
+
+export function getItemCredits(id: string): Result<Credit[], LoadError> {
+  if (!creditsStage.resolved) return err(loading("credits"));
+  return ok(itemCreditsStore?.[id] ?? []);
```
Gaurav0: getItemMerged() returns not-found for a missing id but getItemCredits() and getItemLayers() do not. Shouldn't we have consistent behavior across all get* functions?

Gaurav0 (changes requested):
1. package.json sideEffects is stale and should now point to catalog.ts
2. docs (README.md and CONTRIBUTING.md) referencing catalog.js should be updated

sources/state/catalog.ts
cdvv7788: Yes. That is part of the next pass (consumers are using `unwrapOr([])`. I am going to get rid of those next.
In the meantime, I changed the functions to return Ok/Err for consistency

Gaurav0 (approved):

PR #456, opened 2026-04-28
refactor(state): convert install-item-metadata + preview-canvas-loading + meta to .ts

cdvv7788:
Mechanical-mostly conversions.

Two semantic upgrades while we're here:

- meta.ts: getItemMetadata (DI dep) and the three sorted-layers functions now return Result<T, LoadError> instead of T | null. Production callers in renderer.js and zip.js patch with .unwrapOr(default).

- preview-canvas-loading.ts: getPreviewCanvasState() returns a discriminated union (PreviewState) instead of an opaque string|null. The 5 reasons for showing/hiding the overlay (rendering, ready, loading-layers, canvas-not-initialized, bootstrap-pending) are now individually nameable. UI copy moves to the consumer with a TS-exhaustive switch.

Verified: type-check, lint, 118 node tests, 433 browser tests, build.


sources/state/meta.js
```diff
@@ -1,188 +0,0 @@
...
- * Sort Layers by zPos
- *
- * @param {string} itemId - The ID of the item to get layers for
- * @param {boolean} standardOnly - If true, only include standard animation layers
- * @returns {Array} Sorted array of layers with layerNum and zPos
- */
-export function getSortedLayers(itemId, standardOnly = false) {
-  const meta = metaDeps.getItemMetadata(itemId);
-  if (!meta) {
-    console.error("Item metadata not found:", itemId);
```
Gaurav0, line 42: This seems to have been removed. Is that safe?

sources/state/meta.js
```diff
@@ -1,188 +0,0 @@
...
- * Split Layers by Animation Type, Then Sort by zPos
- *
- * @param {string} itemId - The ID of the item to get layers for
- * @param {boolean} customOnly - If true, only include custom animation layers
- * @returns {Object} Object with animation names as keys and sorted arrays of layers
- */
-export function getSortedLayersByAnim(itemId, customOnly = false) {
-  const meta = metaDeps.getItemMetadata(itemId);
-  if (!meta) {
-    console.error("Item metadata not found:", itemId);
```
Gaurav0, line 87: This seems to have been removed. is that safe?

Gaurav0 (commented):
Please ensure that we still get errors **in the console** when metadata is not-found.

sources/state/meta.js
cdvv7788: WDYM? The errors or the code? The code was moved to the ts file. I just readded the logs there. In my tests against the preview everything still works, plus all the tests are passing (which are more strict than ever). If the coverage was decent we should be ok.

Gaurav0 (approved):

PR #457, opened 2026-04-29
refactor(state): convert state.js to .ts

cdvv7788:
Same old same old. Mostly focused on getting it to work, but the state is touched everywhere. New `unwrapOr` to handle post migration.


sources/state/state.ts
```diff
@@ -0,0 +1,290 @@
...
+ * narrowed at each access. Modeling the DI return as `Partial<ItemMerged>`
+ * matches that style and keeps test stubs (which supply only the fields they
+ * exercise) typeable from JS.
+ */
+type MetadataView = Partial<ItemMerged>;
+
+type ZipMode = { isRunning: boolean };
+
+/** Global application state. Mutated in place; Mithril views observe via redraw. */
+export type State = {
```
Gaurav0, line 34: I'm curious whether we should convert this to a singleton global class in the future rather than just a plain object and if we could gain strictness from that to prevent accidentally having two state objects.

This is fine for now.

sources/state/state.ts
```diff
@@ -0,0 +1,290 @@
...
+  showTransparencyGrid: boolean;
+  applyTransparencyMask: boolean;
+  matchBodyColorEnabled: boolean;
+  compactDisplay: boolean;
+  customUploadedImage: HTMLImageElement | null;
+  customImageZPos: number;
+  previewCanvasZoomLevel: number;
+  fullSpritesheetCanvasZoomLevel: number;
+  /** True after `main.js` runs the first bootstrap `renderCharacter`. */
+  previewBootstrapRenderDone: boolean;
```
Gaurav0, line 52: Not your issue, but this probably belongs in tranisent state.

sources/state/state.ts
```diff
@@ -0,0 +1,290 @@
...
+  /** Mirrored from `renderCharacter` compositing (see `renderer.js`). */
+  isRenderingCharacter: boolean;
```
Gaurav0, line 54: Not your issue, but this probably belongs in transient state.

sources/state/state.ts
```diff
@@ -0,0 +1,290 @@
+// Global state and state operations
+import m from "mithril";
+import { LICENSE_CONFIG, ANIMATIONS, BODY_TYPES } from "./constants.ts";
+import { syncSelectionsToHash, loadSelectionsFromHash } from "./hash.js";
+import { getItemMerged, type ItemMerged } from "./catalog.ts";
+import { renderCharacter } from "../canvas/renderer.js";
+
+/** A single item selection within a selection group (e.g. body, head, ears). */
+export type Selection = {
```
Gaurav0, line 9: I recommend we should refactor this to a class in the future.

For a follow up PR.

sources/state/state.ts
```diff
@@ -0,0 +1,290 @@
...
+function createDefaultStateDeps(): StateDeps {
+  return {
+    getItemMetadata: (itemId) => getItemMerged(itemId).unwrapOr(null),
+    selectDefaults,
+    redraw: () => m.redraw(),
+    syncSelectionsToHash,
+    renderCharacter,
+    loadSelectionsFromHash,
+    getCanvasRenderer: () =>
+      (window as unknown as { canvasRenderer?: unknown }).canvasRenderer,
```
Gaurav0, line 89: awkward.

Gaurav0 (approved):
Comments are for suggested future follow up.

sources/state/state.ts
cdvv7788: I think we should get rid of singletons in general, moving towards full DI. We have a bunch of god objects floating around, mixed with mutable global state.
A problem for another day tho :)

sources/state/state.ts
cdvv7788: Maybe. For now I am trying to migrate with some degree of correctness. Once that is done, we should be able to refactor more easily.

sources/state/state.ts
cdvv7788: Indeed. Let me check.

cdvv7788: We are monkeypatching `window`, and the dom lib doesn't recognize canvasRenderer as part of the Window type. This is basically an ugly escape hatch. We should be able to get rid of it later.

PR #464, opened 2026-05-01
feat: migrate filters and path from js -> ts

cdvv7788:
Same old same old.

There are some more `undefined | null`, but we are slowly covering more ground.


jrconway3 (approved):
Looks solid, I ran tests and tested it locally, everything looks great to me.

PR #466, opened 2026-05-04
feat: Migrate palettes and hash js->ts

cdvv7788:
A couple more files


cdvv7788:
Low risk, bypassing reviews.

PR #467, opened 2026-05-07
feat: Migrate zip helpers to ts

cdvv7788:


jrconway3 (approved):
I ran tests on this locally, tested look good, and I tested all zips, zip exports look good to me.

PR #471, opened 2026-05-10
* Added Recolors to Head Coverings / Cleanup

jrconway3:
Cleaned up all "Head covering" assets including hoods, bandanas, headbands.

Hoods were the most annoying part, as I had to go through tons and tons of partial pixels throughout the hood assets... but I did those weeks ago.

Also removed the variant/recolor assets now, too.

I forgot about it, but I implemented a minor fix onto my fork to not attempt deploying to github pages if its a fork.


.github/workflows/deploy.yml
```diff
@@ -23,6 +23,7 @@ concurrency:
 jobs:
   # Single deploy job since we're just deploying
   deploy:
+    if: ${{ !github.event.repository.fork }}
```
Gaurav0: I use this to test github pages deploy without risking breaking production. Please remove. You can disable the deploy on your fork through settings if you'd like, I think.

Gaurav0 (changes requested):
Basically looks good to me. Just remove the change to github pages deploy.

.github/workflows/deploy.yml
jrconway3: removed

Gaurav0 (approved):

PR #473, opened 2026-05-12
feat: Migrate renderer js -> ts

cdvv7788:


cdvv7788:
Looks ok, moving forward

PR #484, opened 2026-05-15
Custom image cleanup

cdvv7788:
## Summary

  Two related cleanups on top of the recently-completed TS migration:

  1. **Latent `loadImage(null)` bug.** `runRenderCharacter` was using `getSpritePath(...).unwrapOr(null)` and pushing items with `spritePath: null` on err. Downstream the loader branches on `item.customImage` first and otherwise calls
  `loadImage(item.spritePath!)`. The bang assertion was safe only when the *only* producer of `spritePath: null` was the user-upload branch — but `unwrapOr(null)` introduced a second producer. When `getSpritePath` actually erred
  (`{kind:"missing-layer"}`, `{kind:"missing-bodytype-path"}`, `LoadError`), the err leaked into `loadImage(null)` and surfaced as a generic `Failed to load sprite: null` warn that lost all diagnostic information. Fix: branch on the Result, skip
   the push on err with a specific PathError-shaped warn.

  2. **Conflated renderer / JSON contracts.** `ItemToDraw` had `spritePath: string | null` plus `customImage?: HTMLImageElement` — a discriminated state encoded across two fields with a non-local runtime invariant. `RenderedLayer` (the only
  other type in the chain) was shaped around `JSON.stringify`'s quirks: `Omit<>` based on `ItemToDraw`, with the `HTMLImageElement` left to serialize as `{}` and `fileName: ""` doing double duty as identity + custom-discriminant. Refactor:
  introduce `LayerSource = { kind: "catalog"; spritePath } | { kind: "custom"; image }` as the discriminated union, rename `ItemToDraw` → `DrawCall` (1:1 with `drawImage` calls), drop `RenderedLayer` entirely. Renderer's only export is now
  `drawCalls: DrawCall[]` — the flat draw plan. JSON-specific work (dedup by `(itemId, layerNum)`, `supportedAnimations` aggregation, dropping the `HTMLImageElement`, stripping the `spritesheets/` URL prefix) moves to a new
  `serializeLayersForJson(...)` in `state/json.ts`, called by the two production export paths.

  Also dropped the unused `isCustom: boolean` field on `ItemToDraw` / `DrawCall` — only ever written as `false`, never read.

  ## Changes

  - `sources/canvas/renderer.ts` — `DrawCall` + `LayerSource` types; `drawCalls` export replaces the previous `layers` + `itemsToDraw` pair; `formatPathError` helper; loader uses tagged narrowing on `source.kind`.
  - `sources/state/catalog.ts` — `formatLoadError` extracted next to `LoadError`; reusable by the 12 other getters that return `Result<T, LoadError>`.
  - `sources/state/json.ts` — new `SerializedLayer` type and `serializeLayersForJson` function. `exportStateAsJSON` keeps its opaque `readonly unknown[]` pass-through signature so the existing test (and any ad-hoc callers) is unaffected.
  - `sources/utils/zip-helpers.ts` — `addCharacterJsonAndCredits` now takes `readonly DrawCall[]` and serializes internally.
  - `sources/components/download/Download.ts`, `sources/state/zip.ts` — import `drawCalls` and pass to the typed helpers.
  - Test fixtures + specs updated for the renames.

  ## Wire format change

  `character.json`'s per-layer shape changed. Catalog entry:

  ```diff
   {
     "itemId": "body",
     "variant": "light",
     "zPos": 10,
     "layerNum": 1,
     "yPos": 0,
  -  "isCustom": false,
  -  "fileName": "body/bodies/male/walk/light.png",
  +  "source": { "kind": "catalog", "spritePath": "body/bodies/male/walk/light.png" },
     "supportedAnimations": ["walk", "slash", ...]
   }
  ```

  Custom-upload entry:

  ```diff
   {
     "itemId": "custom-upload",
     "variant": null,
     "zPos": 87,
     "layerNum": 0,
     "yPos": 0,
  -  "isCustom": false,
  -  "customImage": {},
  -  "fileName": "",
  +  "source": { "kind": "custom" },
     "supportedAnimations": ["walk", "slash", ...]
   }
  ```

  `spritePath` content is identical, just relocated under `source`. `importStateFromJSON` doesn't read the `layers` field at all, and no in-repo code consumes the old fields. External tools parsing `character.json` for `fileName` / `isCustom`
  would see them disappear.

### DIFF
--- a/sources/canvas/renderer.ts
+++ b/sources/canvas/renderer.ts
@@ -4,7 +4,7 @@
 import { ok, err, type Result } from "neverthrow";
 import { loadImage, loadImagesInParallel } from "./load-image.ts";
 import type { LoadedImage } from "./load-image.ts";
-import { getSpritePath } from "../state/path.ts";
+import { getSpritePath, type PathError } from "../state/path.ts";
 import { getImageToDraw } from "./palette-recolor.ts";
 import { getMultiRecolors } from "../state/palettes.ts";
 import { get2DContext, getZPos } from "./canvas-utils.ts";
@@ -22,7 +22,11 @@ import {
 } from "./preview-animation.ts";
 import { getSortedLayersByAnim } from "../state/meta.ts";
 import type { AnimationLayer } from "../state/meta.ts";
-import { catalogReady, getItemMerged } from "../state/catalog.ts";
+import {
+  catalogReady,
+  formatLoadError,
+  getItemMerged,
+} from "../state/catalog.ts";
 import m from "mithril";
 import { debugWarn } from "../utils/debug.ts";
 import type { Selections } from "../state/state.ts";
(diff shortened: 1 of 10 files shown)


Gaurav0 (commented):
We need to add more test coverage so that bugs like don't appear.

Gaurav0 (approved):

jrconway3 (approved):
I've tested locally and everything looks good. I don't see anything notable in the code changes either.

PR #488, opened 2026-05-19
Refactor catalog

cdvv7788:
This shouldn't break anything. I am changing the catalog from a singleton to a dependency. This allows us to reason about what pieces we need in each component and to control concurrent mutations better by limiting accessors.
Once everything has been migrated, I will clean up the singleton.

### DIFF
--- a/sources/components/App.ts
+++ b/sources/components/App.ts
@@ -2,11 +2,21 @@
 import m from "mithril";
 import { state } from "../state/state.ts";
 import { syncSelectionsToHash } from "../state/hash.ts";
+import type { CatalogReader } from "../state/catalog.ts";
 import { Download } from "./download/Download.ts";
 import { FiltersPanel } from "./FiltersPanel.ts";
 import { Credits } from "./download/Credits.ts";
 import { AdvancedTools } from "./advanced/AdvancedTools.ts";
 import { renderCharacter } from "../canvas/renderer.ts";
+
+/**
+ * App is the composition root for catalog DI. main.ts mounts it with the
+ * `defaultCatalog` instance; App threads narrow slices down to children that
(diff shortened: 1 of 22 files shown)


Gaurav0 (changes requested):
Seeing a huge performance regression on page load as measured by lighthouse.

cdvv7788:
> Seeing a huge performance regression on page load as measured by lighthouse.

Please post the differences next time, better to have a concrete number.

This is my measurement against this branch:
image

This is against master:
image

What should I be looking for here specifically?

I ran `npm run build` -> `npm run preview` and then ran a lighthouse test in the chrome devtools. Everything locally (both branches).

jrconway3 (approved):
In terms of actually breaking anything, everything looks good to me. Tests still work, I can't see anything notably wrong with the interface. I ran my common tests and they all work.

Gaurav0:
screenshot-master

screenshot-pr

Gaurav0:
I'm not sure what to look for, honestly. I don't see anything in your code that should cause such a major impact on performance one way or another.

I think seeing a performance improvement locally for you seems suspect, unless you did any performance tuning.

cdvv7788:
> I'm not sure what to look for, honestly. I don't see anything in your code that should cause such a major impact on performance one way or another.

I would guess part of the issue is testing gh vs the preview. They may have different performance profiles that can affect the readings.

Gaurav0:
Yes. Comparing against https://deploy-preview-487--liberatedpixelcup.netlify.app/#sex=male&body=Body_Color_light&head=Human_Male_light&expression=Neutral_light produced similar results. My guess is that deploy-preview injected stuff is the problem.

Gaurav0 (approved):

PR #490, opened 2026-05-25
Fixed Recolors Issues on #423 and #441

### LINKED ISSUE #423: Recolors Not Setting Default Secondary Color Correctly
When a secondary color is enabled, for some reason it keeps defaulting to the absolute default color:
Image

It should be defaulting to whatever the default color is set to at that time, on that asset. In this case, its brown, but its defaulting to recoloring back to overall "default" asset instead of the type of material, in this case, white for cloth.

### LINKED ISSUE #441: Recolors not working if I unselect `match body colors`
## To reproduce

  - Add a head and paint it in any color
  - Unselect `match body colors`
  - Select a different head and try to recolor

The default behaviour is also confusing. It is `match body colors` but it is the body that is matching the head here.

jrconway3:
Resolves #423 
Resolves #441 

This is mostly for #423 but I also made sure to fix #441 as well.

Only minor thing I notice here: If you turn off match body colors, switch the non-body asset to a different color, then go back to match body color again, it doesn't re-adjust the assets to match the body color.

Also, if a custom color is provided as the default for a secondary color, it now gets its own unique "Custom" section in the recolors list so you can revert back to this color.

Most of the work was done by Copilot/Claude, but Copilot was going so far out of its way to fix the issue adding so much unneeded extra code that I killed that, fixed parts of it myself, then had Claude clean up the rest and fix up the tests.


sources/components/tree/PaletteSelectModal.ts
```diff
@@ -177,18 +189,22 @@ function renderModal(
                     ),
```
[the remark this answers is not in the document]
jrconway3: Yes, that's why only versions that actually exist can be included and why I added a new "custom" already.

sources/components/tree/PaletteSelectModal.ts
```diff
@@ -177,18 +189,22 @@ function renderModal(
...
                   ],
                 ),
-                isExpanded
+                isExpanded || version === CUSTOM_VERSION
                   ? m("div.variants-container.is-flex.is-flex-wrap-wrap", [
```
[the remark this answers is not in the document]
jrconway3: This shouldn't be here, all it does is mean you can close the custom entry, making it pointless.

jrconway3:
I'm not getting these lint issues locally... best I can think is maybe I'm not up-to-date with npm? But that doesn't make much sense either. I guess I'll try it...

EDIT: That's because its type-check rather than lint, which isn't the same thing, but the CI entry is called "lint".

sources/state/state.ts
```diff
@@ -290,7 +290,7 @@ export function selectItem(
         recolor: parentRecolor,
         name: parentRecolor
           ? `${meta.name} (${parentRecolor.replaceAll("_", " ")})`
-          : meta.name,
+          : (meta.name ?? ""),
```
Gaurav0: I'm not sure we fix this in this way. Shouldn't we really should be fixed by setting the return type of getItemMetaData to assert or require a name? @cdvv7788 ?

sources/components/tree/PaletteSelectModal.ts
```diff
@@ -177,18 +189,22 @@ function renderModal(
...
                       ...Object.entries(recolors).map(([palette, colors]) => {
                         const gradient = colors.slice().reverse();
                         const key =
-                          (material !== opt.material ? material + "." : "") +
-                          (version !== opt.default ? version + "." : "") +
-                          palette;
+                          version === CUSTOM_VERSION
```
Gaurav0: the nested ternary operators are hard to read. Please refactor.

sources/state/hash.ts
```diff
@@ -394,8 +392,16 @@ export function loadSelectionsFromHash(hashString: string | null = null): void {
           recolorToMatch,
           subItem.subId,
         );
+        matchedSubItem = true;
+        break;
       }
     }
+
+    if (!matchedSubItem) {
+      debugWarn(
```
Gaurav0, line 401: Good. Moving this should cut down on the noise.

sources/state/palettes.ts
```diff
@@ -297,10 +306,15 @@ export function getPaletteOptions(
...
         selectionColor: selectedColor,
+        sourceColors: color.source ?? null,
         colors:
-          material !== undefined
-            ? getTargetPalette(material, `${version}.${recolor}`).unwrapOr(null)
-            : null,
+          selectedColor === "source" || (!selectedColor && color.source)
+            ? (color.source ?? null)
+            : material !== undefined
+              ? getTargetPalette(material, `${version}.${recolor}`).unwrapOr(
```
Gaurav0: nested ternary operators make for hard to read code. Please refactor.

Gaurav0 (commented):
Mostly okay with this. Small nitpicks, lets not nest ternary operators.

The major question is about the TypeScript fix. I think getItemMetaData should probably always return a name in its type definition. How do we handle this elsewhere?

jrconway3:
> Mostly okay with this. Small nitpicks, lets not nest ternary operators.
> 
> The major question is about the TypeScript fix. I think getItemMetaData should probably always return a name in its type definition. How do we handle this elsewhere?

Not sure why I didn't see any notifications for this, but I'll go ahead and review it.

sources/state/state.ts
jrconway3: I don't think meta.name can be null anyway, I traced it back and `name` is just a string (meta can be null, but not meta.name).

If parentRecolor is valid it doesn't even check anyway. Its just stupid overengineering on GPT and/or Claude's part and I didn't carefully check it.

I rewrote it to be this now:
```
        name: `${meta.name}
          ${parentRecolor ? `(${parentRecolor.replaceAll("_", " ")})` : ""}`,
```

That should be fine I think. I'll push this fix once I finish my tests.

sources/state/palettes.ts
jrconway3: Yeah I cleaned it up. Will be pushed shortly.

sources/components/tree/PaletteSelectModal.ts
jrconway3:
```
const key = compilePaletteKey(
  material,
  version,
  palette,
  opt,
);
```

Maybe a little annoying for just one thing, but I don't like the idea of including too much extra logic inside the PaletteSelectModal.

```
/**
 * Compile Palette Key for Recolor Modal
 */
export function compilePaletteKey(
  version: string,
  material: string,
  palette: string,
  opt: PaletteOption,
): string {
  if (version === CUSTOM_VERSION) {
    return CUSTOM_KEY;
  }

  let key = material !== opt.material ? material + "." : "";

  if (version !== opt.default) {
    key += version + ".";
  }

  return key + palette;
}
```

Gaurav0:
Please remove the line as previously requested.

jrconway3:
> Please remove the line as previously requested.

I noticed a similar issue on another PR I made. I think what happened is I mistakenly committed that to my fork's master branch, so whenever I pulled master and merged master I re-introduced it into whatever other branch I had even when it wasn't there originally. Anyway its removed now.

Gaurav0 (approved):

PR #494, opened 2026-05-26
feat: Migrate path to use catalog DI

cdvv7788:
Another day another bit.

There are many catalog specific subsets in these types. Once this refactor is over, I may change the APIs to stop using a broad catalog object, and pass the actual values directly where it makes sense.

### DIFF
--- a/sources/canvas/canvas-utils.ts
+++ b/sources/canvas/canvas-utils.ts
@@ -63,7 +63,7 @@ export function hasContentInRegion(
 }
 
 export function getZPos(
-  catalog: Pick<CatalogReader, "getItemMerged">,
+  catalog: CatalogReader,
   itemId: string,
   layerNum: number = 1,
 ): number {
--- a/sources/canvas/palette-recolor.ts
+++ b/sources/canvas/palette-recolor.ts
@@ -10,10 +10,9 @@ import {
 import { debugLog, debugWarn } from "../utils/debug.ts";
 import { get2DContext } from "./canvas-utils.ts";
 import { getItemLite } from "../state/catalog.ts";
-import type { ItemMerged } from "../state/catalog.ts";
+import type { CatalogReader, ItemMerged } from "../state/catalog.ts";
 import { state } from "../state/state.ts";
 import { getLayersToLoad } from "../state/meta.ts";
-import type { LayersToLoadCatalog } from "../state/meta.ts";
 import { getPalettesFromMeta, getTargetPalette } from "../state/palettes.ts";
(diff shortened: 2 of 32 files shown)


Gaurav0 (commented):
How much type safety we are buying by using so many types instead of just using CatalogReader? i think we are overengineering here.

cdvv7788:
> How much type safety we are buying by using so many types instead of just using CatalogReader? i think we are overengineering here.

It is possible. I want to get rid of the singleton, but there are just too many places. After that I plan to clean up with specific values instead of passing the whole catalog around.
If I keep it with CatalogReader it is harder to know what pieces need what data later. This intermediate complication should smooth out the later part. That is my reasoning at least.

Gaurav0:
> > How much type safety we are buying by using so many types instead of just using CatalogReader? i think we are overengineering here.
> 
> It is possible. I want to get rid of the singleton, but there are just too many places. After that I plan to clean up with specific values instead of passing the whole catalog around. If I keep it with CatalogReader it is harder to know what pieces need what data later. This intermediate complication should smooth out the later part. That is my reasoning at least.

I know you want to get rid of the singleton. DI helps with that, but I don't see how passing specific values does.

cdvv7788:
Well, at the moment we are passing a huge object, and functions only use a small portion of the data there. Once we have DI in place, we can narrow it down further. Instead of passing the catalog down, we can call the functions directly with the values. This will allow us to reduce the amount of functions/modules that need to be `catalog aware`.
For the outer layers this is not practical (too many values), but for the leaves it could be a useful improvement in clarity.

cdvv7788:
With that said, it will be faster to pass the whole catalog without tightening types here. I can figure out the leaves later.

Gaurav0:
> Well, at the moment we are passing a huge object, and functions only use a small portion of the data there. Once we have DI in place, we can narrow it down further. Instead of passing the catalog down, we can call the functions directly with the values. This will allow us to reduce the amount of functions/modules that need to be `catalog aware`. For the outer layers this is not practical (too many values), but for the leaves it could be a useful improvement in clarity.

I think you're overengineering here. "Passing a huge object" doesn't cost anything, it's just a pointer, and being "catalog aware" is pretty much required for everything since you need to know what's loaded. On the other hand, copying a part of an object does cost time and space.

cdvv7788:
ok, I will just pass the whole thing around

Gaurav0 (approved):

Gaurav0:
Closed in favor of #511

PR #497, opened 2026-06-02
fix playwright install stalled

Gaurav0:


Gaurav0:
It looks like the screen captures are happening at a different width.

Gaurav0:
Turns out it was system fonts not installed in the image.

jrconway3 (approved):
Looks like everything's working to me.

PR #511, opened 2026-06-17
feat: Migrate path to use catalog DI

Gaurav0:
Rebased #494 against latest master

### DIFF
--- a/sources/canvas/canvas-utils.ts
+++ b/sources/canvas/canvas-utils.ts
@@ -63,7 +63,7 @@ export function hasContentInRegion(
 }
 
 export function getZPos(
-  catalog: Pick<CatalogReader, "getItemMerged">,
+  catalog: CatalogReader,
   itemId: string,
   layerNum: number = 1,
 ): number {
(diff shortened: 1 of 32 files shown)


jrconway3 (approved):
It works and doesn't break anything.


I'm not going to decline it, but I will say:

I do NOT like this. Maybe its perfectly fine, but I dislike the idea of passing around this variable everywhere that was perfectly fine being handled via a globals system before.

It just seems to needlessly overcomplicate things and inject this one var into literally EVERYTHING because now EVERYTHING must load this one var.


Maybe I missed something on the prior PR's or whatever, IDK. But to me, loading a helper on each page that loads the catalog data we need just makes more sense, sort of like a helper or services layer. Maybe that's not really a Typescript thing, I'm more used to Laravel for backend infrastructure.

The closest thing we have to a database is the json files metadata. That's effectively our "repository" where we load all data out if that and inject it into the content.


However, this is just my opinion. The current system works, I just think it looks like a complete mess.

Gaurav0:
I'm generally in favor of dependency injection over globals in most cases.

PR #520, opened 2026-07-16
Cleaning the body parts that are attached to the "messy3" hairstyle

kaburagisec:
Screenshot 2026-07-16 165919

When I opened the spritesheet for the "messy3" (https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/tree/master/spritesheets/hair/messy3/adult) hairstyle, I noticed a body part with a light color palette attached to every animation frame (in front/down direction) in the hairstyle.

In this PR, I've already cleaned up the hair from that body part using Aseprite.


Gaurav0 (approved):
LGTM

jrconway3:
Yeah a number of early assets had colors left over from other layers in the past and I fixed some of them, but didn't catch all of them. Looks good, though.

kaburagisec:
Thank You !

PR #527, opened 2026-07-22
Implemented Recolors for Tools + Tools Cleanup

### LINKED ISSUE #342: Split Tool Assets Into Separate Sheets
Tool Assets should be split into separate assets instead of being "variants" of each other.

Ideally they also get recolors added, too.

From this issue:
https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/issues/339

We might want a custom animation order as well, and if we do add custom animation orders, that will only work if they're split up. I checked and it looks like the hammer animation has different missing frames from some of the other tools, indicating they're supposed to have a different animation order (I think).

jrconway3:
- Cleaned up tools to support recolors
- Fixed displaying custom animations for tools
    - Watering restored for Water Can
    - Tool Axe for Axe / Pickaxe
    - Tool Hammer for Hammer
    - Credits Fixes for Assets (all tools are OGA-BY except for Hoe and Shovel)
    - resolves #342
- Recolors for Prosthetics
    - Also Added Idle/Emote
    - Credits Fix for Prosthetics


jrconway3:
I have no idea how to fix this weird Argos issue... I have never seen this issue before.

Gaurav0 (approved):

jrconway3:
I saw that option to support "unsafe actions" but I didn't want to push that myself. That's what Claude recommended in any case.

PR #533, opened 2026-07-28
Fix frame position for "combat_idle" in several helmets

kaburagisec:
The helmet position in the `combat_idle` animation in the upward direction on the `morion`, `nasal`, and `pointed` helmets doesn't move. In this PR, I've fixed this and adjusted the movement to 2 pixels like the other helmet directions.


jrconway3 (approved):
Its probably something I missed because this is when I manually positioned them. Later on I reworked it to support using LPC Tools to position them, but some older assets might be slightly off if I eyed them incorrectly or just forgot.

kaburagisec:
Got it, thank you!

PR #562, opened 2026-08-16
Create a guide for AI agents to follow; Add skills; Documentation updates with an eye to AI agents reading them

Gaurav0:
This is significantly revised. Let me know what you think.

I have been testing it and running it through several models with Cursor, including:

- Composer 2.5 / 100k
- Sonnet 5 / 100k
- Grok 4.6 / 300k
- Opus 5 / 1M

### Summary
Thin AGENTS.md to an always-on index (what the app is, six Never rules, which skill to load). Move procedures into nine Cursor skills and walkthroughs into ARCHITECTURE.md / CONTRIBUTING.md, so a skill edit cannot drift from the file every request loads.

Also: full npm-script and CI tables in CONTRIBUTING.md; runtime wiring in ARCHITECTURE.md; small README / profiling / codecov.yml path fixes.

### Test plan
Docs-only; Argos reports no screenshot changes; all AGENTS.md links resolve.
Phase 4: two unprompted dry-run agents still routed correctly — sources/ → typescript / run-one-spec / coverage; sheet_definitions/ → sheet-definition + validate-site-sources. No restored prose needed.


Gaurav0:
Agent GitHub-access test comment (from AGENTS.md GitHub-section experiment). Safe to delete.

AGENTS.md
```diff
@@ -0,0 +1,239 @@
...
+
+Preliminary agent guide for this repository. Prefer the linked docs when they
+are more specific. Update this file when a practice becomes stable.
+
+## What this repo is
+
+A Vite + Mithril character generator for [LPC](https://lpc.opengameart.org)
+sprites. The UI composes layers from `spritesheets/` using JSON under
+`sheet_definitions/` and `palette_definitions/`. Generated catalog metadata is
+loaded in stages; production code reads it through a `CatalogReader`, not
```
cdvv7788: This reads too much like slop, probably added it after you asked about something related. Useless comment.

AGENTS.md
```diff
@@ -0,0 +1,239 @@
...
+Ignored in the report (do not write tests “for coverage” here): `tests/`,
+`dist/`, `spritesheets/`, `sheet_definitions/`, CSS/SCSS, `vite/` plugins,
+`vite.config.ts`, non-generator `scripts/` (image processing, zip profiling,
+coverage merge, fixture-builder, and similar),
+`sources/performance-profiler.ts`, `sources/utils/debug.ts`, and `testem.cjs`.
+
+When changing tracked production code, add or extend a unit spec so the new
+lines run. Do not commit `coverage/` or `.nyc_output/`.
+
+## GitHub
```
cdvv7788: I don’t think this is useful. If you don’t have your agent with gh and it tries to use other mechanisms that is you and your agent’s problem. This is common knowledge, feels out of place here.

AGENTS.md
```diff
@@ -0,0 +1,239 @@
...
+reach `api.github.com` (CONNECT tunnel 403) and `gh` then reports an invalid
+keyring token / Forbidden. That is a sandbox proxy limit, not GitHub treating
+the client as a bot. Do not retry the same sandboxed command after that
+failure — re-run with those permissions.
+
+**Write (comments, reviews, PRs):** use authenticated `gh` (`gh pr comment`,
+`gh api`, …) with `full_network` or `all` on the first call. Unauthenticated
+POSTs to the API will not work.
+
+## PR checklist (agent)
```
cdvv7788: Why include an (agent) section? This whole file is for agents to consume.
Maybe some of this information should be in the readme or in another doc if it is not there yet, and we reference it from this doc to keep it small?

AGENTS.md
```diff
@@ -0,0 +1,239 @@
...
+
+Keep documentation up to date with the change you are making. Documentation
+includes **all Markdown files** (not only `AGENTS.md`, `README.md`,
+`CONTRIBUTING.md`, and the named guides) and **source code comments**. If
+behavior, commands, paths, or conventions shift, update the relevant docs
+and comments in the same work. Do not leave stale `.js` paths or old Node
+version notes behind.
+
+1. Match existing file style; do not reformat unrelated files.
+2. New code is TypeScript with erasable syntax and explicit `.ts` imports.
```
cdvv7788: We should have this all in extra checks, not only in a md file. Agents love to ignore prompts, specially in longer execution contexts.

cdvv7788 (commented):
Too much information that is not agent specific, it could be more to the point imho

AGENTS.md
Gaurav0: lol it is slop. It's an AI generated summary created when I asked for a template for an AGENTS.md

Let me see if I can make it more useful.

AGENTS.md
Gaurav0: It is very useful to Grok. It kept getting it wrong. That said, it probably belongs in a SKILL, not an AGENTS.md since it is not needed every task.

AGENTS.md
Gaurav0: This will be eventually checked when we can turn off allowJs in tsconfig. In the meantime while the code is in transition to TypeScript this is a useful instruction to an agent.

AGENTS.md
Gaurav0: Humans read this too. lol

We have CI for human error for a reason. Humans are much less likely to read documentation and skip steps than agents. Also CI fails a PR. Sometimes you want a little wiggle room.

That said, there's plenty in CI that my agent wasn't checking locally first. The idea was to save time.

cdvv7788: While we don't have everything hard checked by an action jt should be ok. Once we do have them, we can tell it that no PR will be approved if a check is failing.

AGENTS.md
Gaurav0: On second thought, moved the skill to my cursor user config. If it works automatically for you, we don't need it in the repo.

AGENTS.md
Gaurav0: According to AI it is very useful. It overrides several default assumptions agents are likely to make. Nevertheless, I have revised it several times.

jrconway3:
I had Claude review the skills and docs and it mentioned that the documentation and skills are well defined, even acknowledging that it looked like someone deliberately reported a bunch of issues they noticed agents were struggling with.

A minor issue brought up by Claude: there's many chained cross-references, such as between different skills, and they directly reference headers. The concern is that anchor drift after edits could potentially cause problems, especially since CONTRIBUTING.md is actively edited.


Overall I think this looks solid, just some minor issues Claude is a bit concerned about here. I don't know how well to judge these myself, but skimming through them I don't see anything notable.

PR #566, opened 2026-08-22
Performance profiling skill

Gaurav0:
Speed up compositing on the default WebGL path, then give agents a skill to measure it without DevTools.

## Performance

- Reuse WebGL image and palette textures across recolors (no per-strip allocate on a cold composite).
- First paint blits the live WebGL canvas; LRU snapshots wait until idle (or the next miss). Overlapping renders serialize.
- Boot and hash/selection changes call `renderCharacter` once (`App.onupdate` owns it).
- `?recolor=cpu` applies before first paint so WebGL and CPU can be timed from the same run.

## Performance-profiling skill

New [performance-profiling](https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/blob/83e229c062c782814c612d58521f4b5dd4981c31/.agents/skills/performance-profiling/SKILL.md) — agents run `npm run profile:app` / `diff:app-profile` instead of asking the user to open DevTools. Default `--recolor both`; real GPU via `--headed --channel chrome`. Per-step `renderCharacter` phases land on `profiler.snapshot()`. See [PERFORMANCE_PROFILING.md](https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/blob/83e229c062c782814c612d58521f4b5dd4981c31/PERFORMANCE_PROFILING.md).

Also converts the canvas suite to TypeScript with dest-pixel / call-count / work-budget locks, and adds a local patch-miss reporter so coverage holes do not fail Test browsers (`codecov/patch` stays the gate).

## Test plan

- [x] Browser canvas specs (composite, call-count, work budget, deferred recolor)
- [x] WebGL and CPU visual check (user; `?recolor=cpu`)
- [x] `npm run profile:app -- --recolor both` (optional `--headed --channel chrome`)


.agents/skills/performance-profiling/SKILL.md
```diff
@@ -0,0 +1,109 @@
...
+  Run the headless app and ZIP profilers and interpret the JSON the way a
+  human would from window.profiler.report() and the ZIP console table. Use
+  when editing sources/canvas/, load-image, renderer, zip export, zip-helpers,
+  performance-profiler, or when the user mentions performance, profiling,
+  slow render, FPS, or ZIP export timing.
+---
+
+# Performance profiling
+
+Do not ask the user to open DevTools. Run the matching script, then read the
```
cdvv7788, line 13: It uses the same baseline underneath?

.github/workflows/ci.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v6
         with:
-          fetch-depth: 2
+          fetch-depth: 0
```
cdvv7788, line 21: 👍

cdvv7788 (approved):
It is quite a big change, but I see a lot of them in coverage/tests. Looks pretty handy :)

.agents/skills/performance-profiling/SKILL.md
Gaurav0: See the workflow section:

```bash
npm run profile:app:baseline
# …make the change…
npm run profile:app
npm run diff:app-profile -- tmp/baseline-app-profile.json tmp/app-profile.json
```

PR #569, opened 2026-08-24
feat: Migrate current selections to use a model

cdvv7788:
Split catalog access into separate reader/writer capabilities and piloted narrow UI models.

  - Removed shared catalog singleton behavior so catalog instances remain isolated.
  - Restricted metadata loading and fixture setup to CatalogWriter; application code receives only CatalogReader.
  - Simplified writer registration APIs to accept metadata payloads directly.
  - Removed legacy catalog reset behavior and preserved stable readiness promises.
  - Added a composition-root-owned application model provider.
  - Migrated CurrentSelections from catalog/state attrs to a render-ready model with narrow commands.
  - Preserved the component’s existing behavior and practical per-redraw performance profile.
  - Established the path for removing catalog and state from UI component signatures incrementally.
  
  Actually relevant files
  ---
  ```
    Core catalog/DI changes:

  - sources/state/catalog.ts — separate reader/writer capabilities, isolated stores, readiness lifecycle, simplified registration API.
  - sources/install-item-metadata.ts — fire-and-forget loading through the writer.
  - sources/main.ts — composition-root wiring.

  UI model pilot:

  - sources/models/application.ts — application-owned lazy model providers.
  - sources/models/current-selections.ts — moved Current Selections derivation and commands.
  - sources/components/App.ts — passes the lazy provider.
  - sources/components/FiltersPanel.ts — forwards the provider without evaluating it.
  - sources/components/selections/CurrentSelections.ts — presentational component that evaluates the provider only when rendered.
  ```


Gaurav0:
## PR review: [#569 — feat: Migrate current selections to use a model](https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/pull/569)

**Commit reviewed:** [8f4316a7e8](https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/commit/8f4316a7e8a5af2c3cce87c873c54681048a8460)

---

### Overall

The core refactor is sound and I'd approve it after a couple of fixes. The reader/writer split is genuine rather than type-only — `createCatalog()` now builds two distinct runtime objects over the same closure, and [tests/install-item-metadata_spec.js](https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/blob/8f4316a7e8a5af2c3cce87c873c54681048a8460/tests/install-item-metadata_spec.js) asserts the reader has no `registerItemMetadata` property, so the capability boundary is actually tested. Removing `resetForTests()` and the module-global `metadataLoadStarted` flag is the right call: both were global-state workarounds that stopped making sense once catalogs became per-instance. Making the stages `const` is a quiet improvement, since `catalog.ready.*` promises are now stable for the life of an instance instead of being swapped out by a reset. And `createLoadedCatalog()` keeps the writer entirely inside the module, so production bootstrap can't reach it even by accident. CI is green with 100% patch coverage.

Two things need fixing before merge, and the rest are smaller notes.

---

### Blocking: the model is built eagerly, but the component was lazy

[App.view (sources/components/App.ts#L66-L74)](https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/blob/8f4316a7e8a5af2c3cce87c873c54681048a8460/sources/components/App.ts#L66-L74) calls `models.currentSelections()` unconditionally:

```typescript
const { catalog, state, models } = vnode.attrs;
return m("div", [
  m(Download, { catalog, state }),
  m(FiltersPanel, {
    catalog,
    state,
    currentSelections: models.currentSelections(),
  }),
```

[FiltersPanel](https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/blob/8f4316a7e8a5af2c3cce87c873c54681048a8460/sources/components/FiltersPanel.ts) wraps its children in a [CollapsibleSection (sources/components/CollapsibleSection.ts#L53-L54)](https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/blob/8f4316a7e8a5af2c3cce87c873c54681048a8460/sources/components/CollapsibleSection.ts#L53-L54), which drops them entirely when collapsed:

```typescript
!isCollapsed &&
  m("div", { class: "collapsible-content" }, vnode.children),
```

So previously, collapsing "Filters" meant `CurrentSelections.view` was never called and none of the compatibility-check work happened. Now the model — two filter predicates, a `getItemMerged` object spread, a license `Set`, and tooltip string building, all per selection — is computed on every redraw whether or not the panel is visible. That directly contradicts the PR bullet claiming the per-redraw performance profile is preserved. It isn't per animation frame (nothing redraws Mithril from the rAF loop), so the magnitude is modest, but it's pure waste and easy to avoid.

The fix is to pass the provider rather than the value: either `model: () => CurrentSelectionsModel` invoked inside the leaf's view, or hand `models` down and let `CurrentSelections` call its own slice. Either restores laziness.

---

### Blocking: duplicated words in ARCHITECTURE.md

[ARCHITECTURE.md#L51-L54](https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/blob/8f4316a7e8a5af2c3cce87c873c54681048a8460/ARCHITECTURE.md#L51-L54):

> sources/install-item-metadata.ts does a parallel `import()` of all five chunks and calls the matching **the matching writer** `register*Metadata` method as each one lands, so readiness is **staged** rather than all-or-nothing.

"calls the matching / the matching writer" — the replacement was pasted without removing the tail of the old sentence. The new line also runs well past the ~80-column wrapping the rest of that file uses.

---

### Design note on how this scales

[FiltersPanel](https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/blob/8f4316a7e8a5af2c3cce87c873c54681048a8460/sources/components/FiltersPanel.ts) gained a `currentSelections` attr purely to forward it one level down. Since the PR explicitly sets up "the path for removing catalog and state from UI component signatures incrementally," it's worth deciding now whether every future model adds an attr to every intermediate component. Threading `models: ApplicationModels` (or a scoped sub-bag) instead keeps intermediate signatures stable and composes naturally with the laziness fix above.

Relatedly, [ApplicationModels.currentSelections() (sources/models/application.ts)](https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/blob/8f4316a7e8a5af2c3cce87c873c54681048a8460/sources/models/application.ts) reads like an accessor but is a factory that rebuilds the model on each call. Naming it `createCurrentSelections()`, or typing the field as `() => CurrentSelectionsModel`, would make the per-call cost visible at the call site.

---

### Smaller things

The `"License info loading…"` branch in [sources/models/current-selections.ts](https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/blob/8f4316a7e8a5af2c3cce87c873c54681048a8460/sources/models/current-selections.ts) is dead code carried over from the original: `tooltip` is `undefined` whenever `!creditsReady`, so that string can never be rendered. Pre-existing, but this migration was the natural moment to drop it.

The "Module roles" section in [ARCHITECTURE.md](https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/blob/8f4316a7e8a5af2c3cce87c873c54681048a8460/ARCHITECTURE.md) has tables for `sources/state/`, `sources/canvas/`, and `sources/components/`, but the new top-level [sources/models/](https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/tree/8f4316a7e8a5af2c3cce87c873c54681048a8460/sources/models) directory has no entry. The prose at [ARCHITECTURE.md#L193-L196](https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/blob/8f4316a7e8a5af2c3cce87c873c54681048a8460/ARCHITECTURE.md#L193-L196) mentions the model graph, but someone scanning module roles won't find it. [AGENTS.md](https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/blob/8f4316a7e8a5af2c3cce87c873c54681048a8460/AGENTS.md) asks that the change update whatever doc it makes stale.

The rationale comment that used to sit on the load call in [sources/main.ts](https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/blob/8f4316a7e8a5af2c3cce87c873c54681048a8460/sources/main.ts) — that fetches start as the entry module runs so download and parse overlap HTML parse — was deleted along with the call site and isn't carried into the JSDoc on [createLoadedCatalog() (sources/install-item-metadata.ts#L97-L105)](https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/blob/8f4316a7e8a5af2c3cce87c873c54681048a8460/sources/install-item-metadata.ts#L97-L105). The behavior actually got better (loading now starts at [sources/main.ts#L83](https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/blob/8f4316a7e8a5af2c3cce87c873c54681048a8460/sources/main.ts#L83) instead of after line 106), but the constraint that it must stay early is now undocumented, so a future reorder could silently regress it.

Finally, [loadCatalogFromFixtures (sources/state/catalog.ts)](https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/blob/8f4316a7e8a5af2c3cce87c873c54681048a8460/sources/state/catalog.ts) no longer resets, which means re-seeding the same writer overwrites the stores but can't move a resolved stage back to unresolved. That's the better pattern, but nothing documents it — a spec that re-seeds expecting a fresh "loading" catalog would silently see resolved stages instead. A line in the [catalog skill Tests section (.cursor/skills/catalog/SKILL.md)](https://github.com/LiberatedPixelCup/Universal-LPC-Spritesheet-Character-Generator/blob/8f4316a7e8a5af2c3cce87c873c54681048a8460/.cursor/skills/catalog/SKILL.md) saying "create a new catalog rather than re-seeding" would close that gap.

---

> **Note:** This review was created by Cursor using the Claude Opus 5 model.

Gaurav0 (changes requested):
I've let AI review your PR. See the comment it posted on my behalf.

Also, please break down your single huge commit into many logical, atomic commits so that I can review it myself. AI can help you do this.

Gaurav0:
See https://github.com/Gaurav0/Universal-LPC-Spritesheet-Character-Generator/pull/4#pullrequestreview-5013822360

Gaurav0 (dismissed):

Gaurav0 (approved):

